PR #11, opened 2021-12-10
8278420: C2: assert(!n->is_Store() && !n->is_LoadStore()) failed: no node with a side effect

chhagedorn:
The test case fails with the assertion when an actual unreachable store node with only uses outside of the loop is tried to be sunk out of a dead loop in split-if. This is quite an edge case in which C2 is not able to remove the inner loop but the store for `iFldArr2` inside this loop dies due to improved type information after peeling. This removes some memory phis as well and leaves the store `iFldArr1` with only outside the loop uses. A more detailed explanation how we end up in this situation is shown in the comments of the test case.

This suggests that the assertion is too strong. I propose to relax the assertion and bail out if we are trying to sink a store node. However, I don't think that we will reach this code with `LoadStore` nodes as they have other memory outputs inside a loop, preventing to reach this assertion code.

Thanks,
Christian

---------
### Progress
- [ ] Change must be properly reviewed

### Issue
 * [JDK-8278420](https://bugs.openjdk.java.net/browse/JDK-8278420): C2: assert(!n->is_Store() && !n->is_LoadStore()) failed: no node with a side effect

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/11/head:pull/11` \
`$ git checkout pull/11`

`$ git checkout pull/11` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/11/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 11`

`$ git pr show -t 11`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/11.diff">https://git.openjdk.java.net/jdk18/pull/11.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/loopopts.cpp
+++ b/src/hotspot/share/opto/loopopts.cpp
@@ -1495,7 +1495,15 @@ void PhaseIdealLoop::try_sink_out_of_loop(Node* n) {
       Node* early_ctrl = compute_early_ctrl(n, n_ctrl);
       if (n_loop->is_member(get_loop(early_ctrl)) && // check that this one can't be hoisted now
           ctrl_of_all_uses_out_of_loop(n, early_ctrl, n_loop)) { // All uses in outer loops!
-        assert(!n->is_Store() && !n->is_LoadStore(), "no node with a side effect");
+        if (n->is_Store()) {
+          // In some rare cases, a store could only have uses outside the loop. This happens, for example, if a single
+          // inside the loop memory output node dies after loop peeling: A CastII node of a range check gets an updated
+          // type from its iv input in IGVN which could be negative after peeling (e.g. init value of iv = i >= 0,
+          // stride: x < -i). C2, however, is not able to remove the dead loop completely. This only leaves outside the
+          // loop uses which eventually end in uncommon traps. Do not bother with these stores and bail out.
+          return;
+        }
+        assert(!n->is_LoadStore(), "no node with a side effect");
         Node* outer_loop_clone = NULL;
         for (DUIterator_Last jmin, j = n->last_outs(jmin); j >= jmin;) {
           Node* u = n->last_out(j); // Clone private computation per use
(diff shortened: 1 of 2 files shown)


rwestrel:
That doesn't seem right to me. The fact that a CastII dies on a path that doesn't die is a sign, in my opinion, that the graph is broken.
So:
iArrFld2[j] = 8;
has a CastII that becomes top because the loop j <= -1. And the range check for that array access is a predicate. So the loop in which the store to iArrFld2 is is unreachable. That's why the CastII dies. Do I understand correctly?
Isn't the problem that we would need skeleton predicates between the peel iteration and the loop body to catch that the loop is unreachable?

chhagedorn:
Yes, exactly. You're right, this fixes the wrong thing. But using skeleton predicates instead sounds like a good idea! After having looked at it again, I also think that we can indeed use skeleton predicates to handle this corner case. I'm reworking the fix.

chhagedorn:
After digging deeper into this, I found more issues with loop peeling and loop predication by writing additional tests. I basically found all the issues related to skeleton predicates again but this time with an additional peeling before which makes it quite rare and hard to trigger in practice. I think we need the full range of fixes between the loop and the peeled iteration that we have done before:
1. Initialized skeleton predicate as suggested above to remove a dead loop after peeling.
2. Initial skeleton predicates for the initial value and one for the changing stride with unrolling. We could create pre/main/post and then over-unroll the main loop which lets `CastII` nodes die while the main loop is not removed ([JDK-8193130](https://bugs.openjdk.java.net/browse/JDK-8193130), [JDK-8203915](https://bugs.openjdk.java.net/browse/JDK-8203915)).
3. Update data nodes dependent on the predicates before the peeled iteration down to the loop. Because when later unrolling the loop, the cloned data nodes get again the same control input to the predicates before the peeled iteration. This could schedule loads from the main loop even though the main loop is never entered ([JDK-8237859](https://bugs.openjdk.java.net/browse/JDK-8237859) but with peeling before).

In addition, we could also think about adding empty predicates between the loop and the peeled iteration to enable more rounds of loop predication similar to what we are doing for long counted loops. But this is more an idea for an RFE.

There is currently no easy workaround for 1. - 3. and fixing them completely with skeleton predicates is more complex. I don't think it is worth the risk to do this in 18 given how rare 1. - 3. are. On top of that, these are not recent regressions. I will therefore close this PR and re-target this bug to 19/mainline.

PR #19, opened 2021-12-14
8274323: compiler/codegen/aes/TestAESMain.java failed with "Error: invalid offset: -1434443640" after 8273297

smita-kamath:
The failure happens with XX:+DeoptimizeAlot option. I've set reexecute bit and reset the appropriate state for the interpreter to execute the code when deoptimization occurs.

---------
### Progress

### Issue
 * [JDK-8274323](https://bugs.openjdk.java.net/browse/JDK-8274323): compiler/codegen/aes/TestAESMain.java failed with "Error: invalid offset: -1434443640" after 8273297

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/19/head:pull/19` \
`$ git checkout pull/19`

`$ git checkout pull/19` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/19/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 19`

`$ git pr show -t 19`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/19.diff">https://git.openjdk.java.net/jdk18/pull/19.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
+++ b/src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
@@ -4441,14 +4441,21 @@ class StubGenerator: public StubCodeGenerator {
     __ movptr(key, key_mem);
     __ movptr(state, state_mem);
 #endif
-    __ subptr(rsp, 96 * longSize); // Create space on the stack for htbl entries
     __ movptr(subkeyHtbl, subkeyH_mem);
-    __ movptr(avx512_subkeyHtbl, rsp);
     __ movptr(counter, counter_mem);
+// Save rbp and rsp
+    __ push(rbp);
+    __ movq(rbp, rsp);
+// Align stack
+    __ andq(rsp, -64);
+    __ subptr(rsp, 96 * longSize); // Create space on the stack for htbl entries
+    __ movptr(avx512_subkeyHtbl, rsp);
 
     __ aesgcm_encrypt(in, len, ct, out, key, state, subkeyHtbl, avx512_subkeyHtbl, counter);
 
     __ addptr(rsp, 96 * longSize);
+    __ movq(rsp, rbp);
+    __ pop(rbp);
 
     // Restore state before leaving routine
 #ifdef _WIN64
--- a/src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
+++ b/src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
@@ -4453,7 +4453,6 @@ class StubGenerator: public StubCodeGenerator {
 
     __ aesgcm_encrypt(in, len, ct, out, key, state, subkeyHtbl, avx512_subkeyHtbl, counter);
 
-    __ addptr(rsp, 96 * longSize);
     __ movq(rsp, rbp);
     __ pop(rbp);
 
--- a/src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
+++ b/src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
@@ -4414,9 +4414,8 @@ class StubGenerator: public StubCodeGenerator {
     const Register state = c_rarg5;
     const Address subkeyH_mem(rbp, 2 * wordSize);
     const Register subkeyHtbl = r11;
-    const Address avx512_subkeyH_mem(rbp, 3 * wordSize);
     const Register avx512_subkeyHtbl = r13;
-    const Address counter_mem(rbp, 4 * wordSize);
+    const Address counter_mem(rbp, 3 * wordSize);
     const Register counter = r12;
 #else
     const Address key_mem(rbp, 6 * wordSize);
@@ -4425,9 +4424,8 @@ class StubGenerator: public StubCodeGenerator {
     const Register state = r13;
     const Address subkeyH_mem(rbp, 8 * wordSize);
     const Register subkeyHtbl = r14;
-    const Address avx512_subkeyH_mem(rbp, 9 * wordSize);
     const Register avx512_subkeyHtbl = r12;
-    const Address counter_mem(rbp, 10 * wordSize);
+    const Address counter_mem(rbp, 9 * wordSize);
     const Register counter = rsi;
 #endif
     __ enter();
@@ -4443,11 +4441,14 @@ class StubGenerator: public StubCodeGenerator {
     __ movptr(key, key_mem);
     __ movptr(state, state_mem);
 #endif
+    __ subptr(rsp, 96 * longSize); // Create space on the stack for htbl entries
     __ movptr(subkeyHtbl, subkeyH_mem);
-    __ movptr(avx512_subkeyHtbl, avx512_subkeyH_mem);
+    __ movptr(avx512_subkeyHtbl, rsp);
     __ movptr(counter, counter_mem);
 
     __ aesgcm_encrypt(in, len, ct, out, key, state, subkeyHtbl, avx512_subkeyHtbl, counter);
+
+    __ addptr(rsp, 96 * longSize);
 
     // Restore state before leaving routine
 #ifdef _WIN64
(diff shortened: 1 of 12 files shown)


sviswa7 (approved):
Looks good to me.

sviswa7:
@vnkozlov Could you please also review this patch?

vnkozlov (changes requested):
Yes, we need to reexecute because code could be deoptimized during `new_array()` allocation.
But why we allocate this temp array in Java heap? Why not on stack in stub code?

Also I noticed next return from intrinsics code could be moved up before we generate new nodes in graph: `if (Matcher::htbl_entries == -1) return false;`

smita-kamath:
Thanks for your comments Vladimir. I will make the change and move the check.

smita-kamath:
@vnkozlov Allocating the array in the stub will cause few changes on x86-64 side as well as change in aarch64 stubGenerator code as subkeyHtbl_48_entries will no longer be passed as an argument.  
Do let me know if you think it is okay to proceed with these changes. Thank you.

vnkozlov:
> @vnkozlov Allocating the array in the stub will cause few changes on x86-64 side as well as change in aarch64 stubGenerator code as subkeyHtbl_48_entries will no longer be passed as an argument.
> Do let me know if you think it is okay to proceed with these changes. Thank you.

@smita-kamath Thank you for looking on it. Okay, let proceed with your current fix for JDK 18. File RFE for JDK 19 to rework the code.

Meanwhile I will test this fix.

vnkozlov:
Unfortunately I hit assert during CTW testing on Windows-x64 when compiling `com.sun.crypto.provider.GaloisCounterMode::implGCMCrypt`.

```
#  Internal Error (t:\workspace\open\src\hotspot\share\opto\graphKit.cpp:250), pid=41320, tid=43308
#  assert(ex_map->jvms()->same_calls_as(_exceptions->jvms())) failed: all collected exceptions must come from the same place

Current CompileTask:
C2:   7834 2950    b  4       com.sun.crypto.provider.GaloisCounterMode::implGCMCrypt (100 bytes)

Stack: [0x000000c86a600000,0x000000c86a700000]
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
V  [jvm.dll+0xbb90d1]  os::platform_print_native_stack+0xf1  (os_windows_x86.cpp:235)
V  [jvm.dll+0xdf904e]  VMError::report+0x101e  (vmError.cpp:828)
V  [jvm.dll+0xdfaa4e]  VMError::report_and_die+0x7fe  (vmError.cpp:1656)
V  [jvm.dll+0xdfb1d4]  VMError::report_and_die+0x64  (vmError.cpp:1437)
V  [jvm.dll+0x536f47]  report_vm_error+0xb7  (debug.cpp:280)
V  [jvm.dll+0x6da439]  GraphKit::add_exception_states_from+0x119  (graphKit.cpp:286)
V  [jvm.dll+0x414375]  PredicatedIntrinsicGenerator::generate+0x7f5  (callGenerator.cpp:1358)
V  [jvm.dll+0x5c3b62]  Parse::do_call+0x9c2  (doCall.cpp:651)
V  [jvm.dll+0xbe3745]  Parse::do_one_bytecode+0x32b5  (parse2.cpp:2704)
V  [jvm.dll+0xbd5ae7]  Parse::do_one_block+0x437  (parse1.cpp:1557)
V  [jvm.dll+0xbd463c]  Parse::do_all_blocks+0x5cc  (parse1.cpp:710)
V  [jvm.dll+0xbd0c3d]  Parse::Parse+0xc1d  (parse1.cpp:616)
V  [jvm.dll+0x413a15]  ParseGenerator::generate+0xa5  (callGenerator.cpp:103)
V  [jvm.dll+0x4eba90]  Compile::Compile+0x1110  (compile.cpp:714)
```

The call stack has `PredicatedIntrinsicGenerator` so it seems related to changes.

I got replay file and will try to reproduce it tomorrow.

dean-long:
Vladimir, the assert sounds like JDK-﻿﻿6868269.  Is there an earlier null check that also needs the reexecute flag?

smita-kamath:
@vnkozlov Since, array allocation on heap is causing issues, would it be better if I update the code to have the temp array allocation on stack? The updated code will eliminate the need for re-execution.
I have the change ready and can push it if you think it is alright to do so. Please do let me know your thoughts. Thanks.

vnkozlov:
@dean-long,  there is null check in predicate code:
https://github.com/openjdk/jdk/blob/master/src/hotspot/share/opto/callGenerator.cpp#L1268

This is first time we use `PreserveReexecuteState` for an intrinsic with predicate. But I am actually not sure if the failure I see is caused by this changes or existing bug. But it causes concern.

@smita-kamath yes, please prepare new changes with stack allocation and I will test it.

vnkozlov:
Note, we can't change code in `PredicatedIntrinsicGenerator` because it is used by all intrinsics with predicate and non do re-execution.

dean-long:
Based on past bugs, it looks like to use new_array here, we need to pass /*deoptimize_on_exception=*/true as the optional 5th argument, just like inline_native_clone does.

src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
```diff
@@ -4443,12 +4441,15 @@ class StubGenerator: public StubCodeGenerator {
     __ movptr(key, key_mem);
     __ movptr(state, state_mem);
 #endif
+    __ subptr(rsp, 96 * longSize); // Create space on the stack for htbl entries
```
vnkozlov: Is this aligned correctly? Or alignment does not matter?

src/hotspot/share/opto/library_call.cpp
```diff
@@ -6763,86 +6763,71 @@ bool LibraryCallKit::inline_galoisCounterMode_AESCrypt() {
...
-  Node* in_start = in;
-  Node* ct_start = ct;
-  Node* out_start = out;
-  if (inOfs != NULL || ctOfs != NULL || outOfs != NULL) {
-    assert(inOfs != NULL && ctOfs != NULL && outOfs != NULL, "");
-    in_start = array_element_address(in, inOfs, T_BYTE);
-    ct_start = array_element_address(ct, ctOfs, T_BYTE);
-    out_start = array_element_address(out, outOfs, T_BYTE);
-  }
+    // (1) in, ct and out are arrays.
```
vnkozlov: You need to restore indent.

vnkozlov (commented):
I have few comments. I will start testing.

src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
smita-kamath: Thanks, Vladimir. I've addressed both your comments in the latest update.

src/hotspot/cpu/x86/stubGenerator_x86_64.cpp
```diff
@@ -4444,11 +4442,21 @@ class StubGenerator: public StubCodeGenerator {
...
+    __ push(rbp);
+    __ movq(rbp, rsp);
+// Align stack
+    __ andq(rsp, -64);
+    __ subptr(rsp, 96 * longSize); // Create space on the stack for htbl entries
+    __ movptr(avx512_subkeyHtbl, rsp);
 
     __ aesgcm_encrypt(in, len, ct, out, key, state, subkeyHtbl, avx512_subkeyHtbl, counter);
 
+    __ addptr(rsp, 96 * longSize);
```
vnkozlov: I don't think you need this instruction since you restore `RSP` in the next.
Otherwise looks good.
Testing passed fine.

sviswa7 (approved):
Changes look good to me.

vnkozlov (approved):
I tested without `addptr` instructions and it passed.
It is good for integration.

smita-kamath:
@vnkozlov Thanks a lot for approving this PR. I appreciate your help.

sviswa7:
@vnkozlov @dean-long Thanks a lot for guiding Smita through this fix.

PR #24, opened 2021-12-14
8278508: Enable X86 maskAll instruction pattern for 32 bit JVM.

jatin-bhateja:
- Vector.maskAll was accelerated for AVX-512 target, but  x86 existing backend implementation does not enable maskAll instruction patterns for 32 bit JVM, due to which operations fall backs over replicateB operation which broadcasts the mask value in a vector.
- In some cases after unboxing-boxing optimization this vector eventually reaches to XorVMask which has different operands one held in opmask register and other in vector.

Kindly review and share your feedback.

Best Regards,
Jatin

---------
### Progress

### Issue
 * [JDK-8278508](https://bugs.openjdk.java.net/browse/JDK-8278508): Enable X86 maskAll instruction pattern for 32 bit JVM.

### Reviewers
 * [Vladimir Kozlov](https://openjdk.java.net/census#kvn) (@vnkozlov - **Reviewer**) ⚠️ Review applies to 4fb1ea1b1209d383384953dfcab067064fc60e7f

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/24/head:pull/24` \
`$ git checkout pull/24`

`$ git checkout pull/24` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/24/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 24`

`$ git pr show -t 24`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/24.diff">https://git.openjdk.java.net/jdk18/pull/24.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
@@ -4278,9 +4278,6 @@ void C2_MacroAssembler::vector_maskall_operation(KRegister dst, Register src, in
   if (VM_Version::supports_avx512bw()) {
     if (mask_len > 32) {
       kmovql(dst, src);
-      if (mask_len != 64) {
-        kshiftrql(dst, dst, 64 - mask_len);
-      }
     } else {
       kmovdl(dst, src);
       if (mask_len != 32) {
@@ -4301,6 +4298,5 @@ void C2_MacroAssembler::vector_maskall_operation32(KRegister dst, Register src, 
   assert(VM_Version::supports_avx512bw(), "");
   kmovdl(tmp, src);
   kunpckdql(dst, tmp, tmp);
-  kshiftrql(dst, dst, 64 - mask_len);
 }
 #endif
--- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
@@ -4278,15 +4278,21 @@ void C2_MacroAssembler::vector_maskall_operation(KRegister dst, Register src, in
   if (VM_Version::supports_avx512bw()) {
     if (mask_len > 32) {
       kmovql(dst, src);
-      kshiftrql(dst, dst, 64 - mask_len);
+      if (mask_len != 64) {
+        kshiftrql(dst, dst, 64 - mask_len);
+      }
     } else {
       kmovdl(dst, src);
-      kshiftrdl(dst, dst, 32 - mask_len);
+      if (mask_len != 32) {
+        kshiftrdl(dst, dst, 32 - mask_len);
+      }
     }
   } else {
     assert(mask_len <= 16, "");
     kmovwl(dst, src);
-    kshiftrwl(dst, dst, 16 - mask_len);
+    if (mask_len != 16) {
+      kshiftrwl(dst, dst, 16 - mask_len);
+    }
   }
 }
 
@@ -4294,8 +4300,7 @@ void C2_MacroAssembler::vector_maskall_operation(KRegister dst, Register src, in
 void C2_MacroAssembler::vector_maskall_operation32(KRegister dst, Register src, KRegister tmp, int mask_len) {
   assert(VM_Version::supports_avx512bw(), "");
   kmovdl(tmp, src);
-  kshiftlql(dst, tmp, 32);
-  korql(dst, dst, tmp);
+  kunpckdql(dst, tmp, tmp);
   kshiftrql(dst, dst, 64 - mask_len);
 }
 #endif
--- a/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
+++ b/src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
@@ -4273,3 +4273,29 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
   vector_mask_operation_helper(opc, dst, tmp, masklen);
 }
 #endif
+
+void C2_MacroAssembler::vector_maskall_operation(KRegister dst, Register src, int mask_len) {
+  if (VM_Version::supports_avx512bw()) {
+    if (mask_len > 32) {
+      kmovql(dst, src);
+      kshiftrql(dst, dst, 64 - mask_len);
+    } else {
+      kmovdl(dst, src);
+      kshiftrdl(dst, dst, 32 - mask_len);
+    }
+  } else {
+    assert(mask_len <= 16, "");
+    kmovwl(dst, src);
+    kshiftrwl(dst, dst, 16 - mask_len);
+  }
+}
+
+#ifndef _LP64
+void C2_MacroAssembler::vector_maskall_operation32(KRegister dst, Register src, KRegister tmp, int mask_len) {
+  assert(VM_Version::supports_avx512bw(), "");
+  kmovdl(tmp, src);
+  kshiftlql(dst, tmp, 32);
+  korql(dst, dst, tmp);
+  kshiftrql(dst, dst, 64 - mask_len);
+}
+#endif
--- a/src/hotspot/cpu/x86/x86.ad
+++ b/src/hotspot/cpu/x86/x86.ad
@@ -9451,19 +9451,6 @@ instruct evcmp_masked(kReg dst, vec src1, vec src2, immI8 cond, kReg mask, rRegP
   ins_pipe( pipe_slow );
 %}
 
-instruct mask_all_evexI_imm_LE32(kReg dst, immI cnt, rRegI tmp) %{
-  predicate(Matcher::vector_length(n) <= 32);
-  match(Set dst (MaskAll cnt));
-  effect(TEMP dst, TEMP tmp);
-  format %{ "mask_all_evexI_imm_LE32 $dst, $cnt \t! using $tmp as TEMP" %}
-  ins_encode %{
-    int mask_len = Matcher::vector_length(this);
-    __ movl($tmp$$Register, $cnt$$constant);
-    __ vector_maskall_operation($dst$$KRegister, $tmp$$Register, mask_len);
-  %}
-  ins_pipe( pipe_slow );
-%}
-
 instruct mask_all_evexI_LE32(kReg dst, rRegI src) %{
   predicate(Matcher::vector_length(n) <= 32);
   match(Set dst (MaskAll src));
--- a/src/hotspot/cpu/x86/x86.ad
+++ b/src/hotspot/cpu/x86/x86.ad
@@ -9454,7 +9454,6 @@ instruct evcmp_masked(kReg dst, vec src1, vec src2, immI8 cond, kReg mask, rRegP
 instruct mask_all_evexI_LE32(kReg dst, rRegI src) %{
   predicate(Matcher::vector_length(n) <= 32);
   match(Set dst (MaskAll src));
-  effect(TEMP dst);
   format %{ "mask_all_evexI_LE32 $dst, $src \t" %}
   ins_encode %{
     int mask_len = Matcher::vector_length(this);
--- a/src/hotspot/cpu/x86/x86.ad
+++ b/src/hotspot/cpu/x86/x86.ad
@@ -1827,7 +1827,7 @@ const bool Matcher::match_rule_supported_vector(int opcode, int vlen, BasicType 
       }
       break;
     case Op_MaskAll:
-      if (!is_LP64 || !VM_Version::supports_evex()) {
+      if (!VM_Version::supports_evex()) {
         return false;
       }
       if ((vlen > 16 || is_subword_type(bt)) && !VM_Version::supports_avx512bw()) {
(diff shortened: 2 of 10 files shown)


vnkozlov (commented):
Seems reasonable. Let me test it.

vnkozlov (changes requested):
Regular testing (x64) results are good.

But I also run `jdk/incubator/vector/` tests locally with 32-bit VM (fastdebug). Most tests passed but some hit timeout (run for more than default 2 min):
```
jdk/incubator/vector/Byte512VectorTests.java
jdk/incubator/vector/ByteMaxVectorTests.java
jdk/incubator/vector/VectorReshapeTests.java
```
I assume such vectors are not supported in 32-bit VM and code is slow. Simple solution is to add `/timeout=240` (4 min) to these tests. They passed for me after that.

jatin-bhateja:
> Regular testing (x64) results are good.
> 
> But I also run `jdk/incubator/vector/` tests locally with 32-bit VM (fastdebug). Most tests passed but some hit timeout (run for more than default 2 min):
> 
> ```
> jdk/incubator/vector/Byte512VectorTests.java
> jdk/incubator/vector/ByteMaxVectorTests.java
> jdk/incubator/vector/VectorReshapeTests.java
> ```
> 
> I assume such vectors are not supported in 32-bit VM and code is slow. Simple solution is to add `/timeout=240` (4 min) to these tests. They passed for me after that.

Hi @vnkozlov , x86 32bit target constraints the number of vector registers to 8 (zmm0-zmm7) so above testcases should still be able to intrinsify the operation. I tested above tests on AVX512 target and they executes in lass than 2 mins. On a non-AVX512 target any operation based on Byte512 species is not inline expanded. 

Since the changes in this patch are only related to AVX512 thus my testing went clean with UseAVX=3.

vnkozlov:
I ran these tests with `-XX:-TieredCompilation -Xbatch -XX:UseAVX=3` flags and 32-bit VM on Skylake.

jatin-bhateja:
> ```
> jdk/incubator/vector/VectorReshapeTests.java
> ```

Hi @vnkozlov , I have added explicit time out check as suggested in above 3 testcases.

vnkozlov (approved):
Good. You need second review for this.

jatin-bhateja:
@svisva7 can you kindly be second reviewer.

src/hotspot/cpu/x86/x86_32.ad
```diff
@@ -13847,7 +13847,54 @@ instruct cmpFastUnlock(eFlagsReg cr, eRegP object, eAXRegP box, eRegP tmp ) %{
...
+  match(Set dst (MaskAll cnt));
+  effect(TEMP dst, TEMP tmp, TEMP ktmp);
+  format %{ "mask_all_evex_imm_GT32 $dst, $cnt \t! using $tmp and $ktmp as TEMP" %}
+  ins_encode %{
+    int mask_len = Matcher::vector_length(this);
+    __ movl($tmp$$Register, $cnt$$constant);
+    __ vector_maskall_operation32($dst$$KRegister, $tmp$$Register, $ktmp$$KRegister, mask_len);
+  %}
+  ins_pipe( pipe_slow );
+%}
```
sviswa7: This can be removed. The more generic  mask_all_evexI_GT32 rule will take care of this.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -9451,64 +9451,32 @@ instruct evcmp_masked(kReg dst, vec src1, vec src2, immI8 cond, kReg mask, rRegP
...
   ins_pipe( pipe_slow );
 %}
 
-instruct mask_all_evexL(kReg dst, rRegL src) %{
+instruct mask_all_evexI_LE32(kReg dst, rRegI src) %{
+  predicate(Matcher::vector_length(n) <= 32);
   match(Set dst (MaskAll src));
-  effect(TEMP_DEF dst);
-  format %{ "mask_all_evexL $dst, $src \t! mask all operation" %}
+  effect(TEMP dst);
```
sviswa7: TEMP dst is not needed.

src/hotspot/cpu/x86/x86.ad
```diff
@@ -9451,64 +9451,32 @@ instruct evcmp_masked(kReg dst, vec src1, vec src2, immI8 cond, kReg mask, rRegP
...
-    }
-  %}
-  ins_pipe( pipe_slow );
-%}
-
-instruct mask_all_evexI(kReg dst, rRegI src, rRegL tmp) %{
-  match(Set dst (MaskAll src));
-  effect(TEMP_DEF dst, TEMP tmp);
-  format %{ "mask_all_evexI $dst, $src \t! using $tmp as TEMP" %}
+  effect(TEMP dst, TEMP tmp);
```
sviswa7: TEMP dst is not needed here.

src/hotspot/cpu/x86/x86_32.ad
```diff
@@ -13847,7 +13847,54 @@ instruct cmpFastUnlock(eFlagsReg cr, eRegP object, eAXRegP box, eRegP tmp ) %{
   ins_pipe(pipe_slow);
 %}
 
+instruct mask_all_evexL_LT32(kReg dst, eRegL src) %{
+  predicate(Matcher::vector_length(n) <= 32);
+  match(Set dst (MaskAll src));
+  effect(TEMP dst);
```
sviswa7: TEMP dst is not needed.

src/hotspot/cpu/x86/x86_64.ad
```diff
@@ -13011,6 +13011,43 @@ instruct safePoint_poll_tls(rFlagsReg cr, rRegP poll)
   ins_pipe(ialu_reg_mem);
 %}
 
+instruct mask_all_evexL(kReg dst, rRegL src) %{
+  match(Set dst (MaskAll src));
+  effect(TEMP dst);
```
sviswa7: TEMP dst not needed.

src/hotspot/cpu/x86/x86_64.ad
```diff
@@ -13011,6 +13011,43 @@ instruct safePoint_poll_tls(rFlagsReg cr, rRegP poll)
...
+    int mask_len = Matcher::vector_length(this);
+    __ vector_maskall_operation($dst$$KRegister, $src$$Register, mask_len);
+  %}
+  ins_pipe( pipe_slow );
+%}
+
+instruct mask_all_evexI_GT32(kReg dst, rRegI src, rRegL tmp) %{
+  predicate(Matcher::vector_length(n) > 32);
+  match(Set dst (MaskAll src));
+  effect(TEMP dst, TEMP tmp);
```
sviswa7: TEMP dst not needed.

src/hotspot/cpu/x86/x86_64.ad
```diff
@@ -13011,6 +13011,43 @@ instruct safePoint_poll_tls(rFlagsReg cr, rRegP poll)
...
+    __ movslq($tmp$$Register, $src$$Register);
+    __ vector_maskall_operation($dst$$KRegister, $tmp$$Register, mask_len);
+  %}
+  ins_pipe( pipe_slow );
+%}
+
+instruct mask_all_evex_imm_GT32(kReg dst, immI cnt, rRegL tmp) %{
+  predicate(Matcher::vector_length(n) > 32);
+  match(Set dst (MaskAll cnt));
+  effect(TEMP dst, TEMP tmp);
```
sviswa7: TEMP dst not needed.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4273,3 +4273,29 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
...
+      kmovql(dst, src);
+      kshiftrql(dst, dst, 64 - mask_len);
+    } else {
+      kmovdl(dst, src);
+      kshiftrdl(dst, dst, 32 - mask_len);
+    }
+  } else {
+    assert(mask_len <= 16, "");
+    kmovwl(dst, src);
+    kshiftrwl(dst, dst, 16 - mask_len);
```
sviswa7: If masklen == 16 then kshiftrwl is not needed?

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4273,3 +4273,29 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
   vector_mask_operation_helper(opc, dst, tmp, masklen);
 }
 #endif
+
+void C2_MacroAssembler::vector_maskall_operation(KRegister dst, Register src, int mask_len) {
+  if (VM_Version::supports_avx512bw()) {
+    if (mask_len > 32) {
+      kmovql(dst, src);
+      kshiftrql(dst, dst, 64 - mask_len);
```
sviswa7: Here masklen is 64, so kshiftrql is not needed here?

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4273,3 +4273,29 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
...
+    } else {
+      kmovdl(dst, src);
+      kshiftrdl(dst, dst, 32 - mask_len);
```
sviswa7: If masklen is 32 then kshiftrdl is not needed. Only needed if masklen < 32.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4273,3 +4273,29 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
...
+  }
+}
+
+#ifndef _LP64
+void C2_MacroAssembler::vector_maskall_operation32(KRegister dst, Register src, KRegister tmp, int mask_len) {
+  assert(VM_Version::supports_avx512bw(), "");
+  kmovdl(tmp, src);
+  kshiftlql(dst, tmp, 32);
+  korql(dst, dst, tmp);
+  kshiftrql(dst, dst, 64 - mask_len);
```
sviswa7: Do we need the kshiftrql here? The masklen is 64 here.
You could alternatively use:
   kmovdl dst, src
   kunpckdq dst, dst, dst

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
jatin-bhateja: src operand carry either a 0 (false mask) or 0XFFFFFFFFFFFFFFFF  (true mask) value. Thus a right shift here ensures that destination bits whose size matches with mask_len are set.

I have introduced special case handling when mask_len is  64, 32 and 16.

src/hotspot/cpu/x86/x86.ad
jatin-bhateja: dst is both read and written to in macro assembly routine.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
jatin-bhateja: DONE

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4273,3 +4273,34 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
...
 }
 #endif
+
+void C2_MacroAssembler::vector_maskall_operation(KRegister dst, Register src, int mask_len) {
+  if (VM_Version::supports_avx512bw()) {
+    if (mask_len > 32) {
+      kmovql(dst, src);
+      if (mask_len != 64) {
+        kshiftrql(dst, dst, 64 - mask_len);
+      }
```
sviswa7: We are only supporting vector lengths of power of 2 for x86 (8,16,32,64) for byte vector.  The only case that comes here is mask_len == 64.

src/hotspot/cpu/x86/c2_MacroAssembler_x86.cpp
```diff
@@ -4273,3 +4273,34 @@ void C2_MacroAssembler::vector_mask_operation(int opc, Register dst, XMMRegister
...
+    }
+  }
+}
+
+#ifndef _LP64
+void C2_MacroAssembler::vector_maskall_operation32(KRegister dst, Register src, KRegister tmp, int mask_len) {
+  assert(VM_Version::supports_avx512bw(), "");
+  kmovdl(tmp, src);
+  kunpckdql(dst, tmp, tmp);
+  kshiftrql(dst, dst, 64 - mask_len);
```
sviswa7: We are only supporting vector lengths of power of 2 for x86 (8,16,32,64) for byte vector. The only case that comes here is mask_len == 64 so we dont need kshiftrql.

src/hotspot/cpu/x86/x86.ad
sviswa7: dst both written and read is not the criteria for TEMP dst effect. Only if the src is used after dest is written in the encoding for instruct, TEMP dst is needed.

merykitty: Hi, do we need `TEMP dst` if some `tmp` is used after `dst` is written, as without it `tmp` and `dst` are implied to have non-overlapping lifetime and can lead to conflict. This is a pure question unrelated to the PR. Thank you very much.

sviswa7: Hi @merykitty, yes we would need TEMP dst in that case as well.

jatin-bhateja: TEMP attribute ensures creation of a temporary machine operand which interferes with source operands if their registerclasses are overlapping, a TEMP + DST data flow attribute ensures that DST is not accidentally allocated a SRC register even if SRC is not live beyond that instruction. This shall prevent overriding the src before using its value in the instruction encoding block.

jatin-bhateja:
Hi @sviswa7, your comments addressed. TEMP def was  an artifact of some other change I was doing initially.

sviswa7 (approved):
Patch looks good to me.

PR #25, opened 2021-12-14
8273107: RunThese24H times out with "java.lang.management.ThreadInfo.getLockName()" is null

dcubed-ojdk:
RunThese24H sometimes times out with a couple of error msgs:
- "java.lang.management.ThreadInfo.getLockName()" is null
- ForkJoin common pool thread stuck

The '"java.lang.management.ThreadInfo.getLockName()" is null' error msg was
due to RunThese's use of an older JCK test suite which has since been fixed.

The 'ForkJoin common pool thread stuck' failure mode is likely due to a thread
spending a lot of time in ObjectSynchronizer::monitors_iterate() due to a
VM_ThreadDump::doit() call. I say "likely" because I've never been able to
reproduce this failure mode in testing outside of Mach5. With the Mach5
sightings, all we have are thread dumps and core files and not a live process.

The VM_ThreadDump::doit() call is trying to gather owned monitor information
for all threads in the system. I've seen sightings of this failure mode with > 2000
threads. I've also seen passing runs with > 1.7 million monitors on the in-use list.
Imagine searching a larger in-use list for > 2000 threads. It just doesn't scale.

---------
### Progress

### Issue
 * [JDK-8273107](https://bugs.openjdk.java.net/browse/JDK-8273107): RunThese24H times out with "java.lang.management.ThreadInfo.getLockName()" is null

### Reviewers
 * [Robbin Ehn](https://openjdk.java.net/census#rehn) (@robehn - **Reviewer**) ⚠️ Review applies to d1ca83e7d4a4304c4b502d6adfb551bd8dcf4f11
 * [Coleen Phillimore](https://openjdk.java.net/census#coleenp) (@coleenp - **Reviewer**) ⚠️ Review applies to d1ca83e7d4a4304c4b502d6adfb551bd8dcf4f11
 * [Erik Österlund](https://openjdk.java.net/census#eosterlund) (@fisk - **Reviewer**) ⚠️ Review applies to d1ca83e7d4a4304c4b502d6adfb551bd8dcf4f11

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/25/head:pull/25` \
`$ git checkout pull/25`

`$ git checkout pull/25` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/25/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 25`

`$ git pr show -t 25`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/25.diff">https://git.openjdk.java.net/jdk18/pull/25.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/runtime/synchronizer.cpp
+++ b/src/hotspot/share/runtime/synchronizer.cpp
@@ -1546,8 +1546,8 @@ size_t ObjectSynchronizer::deflate_idle_monitors(ObjectMonitorsHashtable* table)
     ls->print_cr("end deflating: in_use_list stats: ceiling=" SIZE_FORMAT ", count=" SIZE_FORMAT ", max=" SIZE_FORMAT,
                  in_use_list_ceiling(), _in_use_list.count(), _in_use_list.max());
     if (table != nullptr) {
-      ls->print_cr("ObjectMonitorsHashtable: jt_count=" SIZE_FORMAT ", om_count=" SIZE_FORMAT,
-                   table->jt_count(), table->om_count());
+      ls->print_cr("ObjectMonitorsHashtable: key_count=" SIZE_FORMAT ", om_count=" SIZE_FORMAT,
+                   table->key_count(), table->om_count());
     }
   }
 
--- a/src/hotspot/share/runtime/synchronizer.cpp
+++ b/src/hotspot/share/runtime/synchronizer.cpp
@@ -74,17 +74,13 @@ ObjectMonitorsHashtable::~ObjectMonitorsHashtable() {
 
 void ObjectMonitorsHashtable::add_entry(JavaThread* jt, ObjectMonitor* om) {
   ObjectMonitorsHashtable::PtrList* list = get_entry(jt);
-  if (list != nullptr) {
-    // Add to existing list from the hash table:
-    list->add(om);
-    _om_count++;
-  } else {
-    // Create new list, add the ObjectMonitor* to it, and add the list to the hash table:
+  if (list == nullptr) {
+    // Create new list and add it to the hash table:
     list = new (ResourceObj::C_HEAP, mtThread) ObjectMonitorsHashtable::PtrList();
-    list->add(om);
-    _om_count++;
     add_entry(jt, list);
   }
+  list->add(om);  // Add the ObjectMonitor to the list.
+  _om_count++;
 }
 
 bool ObjectMonitorsHashtable::has_entry(JavaThread* jt, ObjectMonitor* om) {
@@ -1030,6 +1026,9 @@ JavaThread* ObjectSynchronizer::get_lock_owner(ThreadsList * t_list, Handle h_ob
 
 // Visitors ...
 
+// Iterate ObjectMonitors where the owner == thread; this does NOT include
+// ObjectMonitors where owner is set to a stack lock address in thread.
+//
 // This version of monitors_iterate() works with the in-use monitor list.
 //
 void ObjectSynchronizer::monitors_iterate(MonitorClosure* closure, JavaThread* thread) {
@@ -1037,6 +1036,8 @@ void ObjectSynchronizer::monitors_iterate(MonitorClosure* closure, JavaThread* t
   while (iter.has_next()) {
     ObjectMonitor* mid = iter.next();
     if (mid->owner() != thread) {
+      // Not owned by the target thread and intentionally skips when owner
+      // is set to a stack lock address in the target thread.
       continue;
     }
     if (!mid->is_being_async_deflated() && mid->object_peek() != NULL) {
@@ -1059,9 +1060,10 @@ void ObjectSynchronizer::monitors_iterate(MonitorClosure* closure,
                                           ObjectMonitorsHashtable::PtrList* list,
                                           JavaThread* thread) {
   typedef LinkedListIterator<ObjectMonitor*> ObjectMonitorIterator;
-  ObjectMonitorIterator iter = ObjectMonitorIterator(list->head());
+  ObjectMonitorIterator iter(list->head());
   while (!iter.is_empty()) {
     ObjectMonitor* mid = *iter.next();
+    // Owner set to a stack lock address in thread should never be seen here:
     assert(mid->owner() == thread, "must be");
     if (!mid->is_being_async_deflated() && mid->object_peek() != NULL) {
       // Only process with closure if the object is set.
@@ -1402,6 +1404,14 @@ void ObjectSynchronizer::chk_for_block_req(JavaThread* current, const char* op_n
 
 // Walk the in-use list and deflate (at most MonitorDeflationMax) idle
 // ObjectMonitors. Returns the number of deflated ObjectMonitors.
+//
+// If table != nullptr, we gather owned ObjectMonitors indexed by the
+// owner in the table. Please note that ObjectMonitors where the owner
+// is set to a stack lock address are NOT associated with the JavaThread
+// that holds that stack lock. All of the current consumers of
+// ObjectMonitorsHashtable info only care about JNI locked monitors and
+// those do not have the owner set to a stack lock address.
+//
 size_t ObjectSynchronizer::deflate_monitor_list(Thread* current, LogStream* ls,
                                                 elapsedTimer* timer_p,
                                                 ObjectMonitorsHashtable* table) {
@@ -1416,11 +1426,14 @@ size_t ObjectSynchronizer::deflate_monitor_list(Thread* current, LogStream* ls,
     if (mid->deflate_monitor()) {
       deflated_count++;
     } else if (table != nullptr) {
-      // The caller is interested in the owned ObjectMonitors. This does not capture
-      // unowned ObjectMonitors that cannot be deflated because of a waiter. Since
-      // deflate_idle_monitors() and deflate_monitor_list() can be called more than
-      // once, we have to make sure the entry has not already been added.
+      // The caller is interested in the owned ObjectMonitors. This does
+      // not include when owner is set to a stack lock address in thread.
+      // This also does not capture unowned ObjectMonitors that cannot be
+      // deflated because of a waiter.
       JavaThread* jt = (JavaThread*)mid->owner();
+      // Since deflate_idle_monitors() and deflate_monitor_list() can be
+      // called more than once, we have to make sure the entry has not
+      // already been added.
       if (jt != nullptr && !table->has_entry(jt, mid)) {
         table->add_entry(jt, mid);
       }
--- a/src/hotspot/share/runtime/synchronizer.cpp
+++ b/src/hotspot/share/runtime/synchronizer.cpp
@@ -57,6 +57,44 @@
 #include "utilities/events.hpp"
 #include "utilities/preserveException.hpp"
 
+class CleanupObjectMonitorsHashtable: StackObj {
+ public:
+  bool do_entry(JavaThread*& key, ObjectMonitorsHashtable::PtrList*& list) {
+    list->clear();  // clear the LinkListNodes
+    delete list;    // then delete the LinkedList
+    return true;
+  }
+};
+
+ObjectMonitorsHashtable::~ObjectMonitorsHashtable() {
+  CleanupObjectMonitorsHashtable cleanup;
+  _ptrs->unlink(&cleanup);  // cleanup the LinkedLists
+  delete _ptrs;             // then delete the hash table
+}
+
+void ObjectMonitorsHashtable::add_entry(JavaThread* jt, ObjectMonitor* om) {
+  ObjectMonitorsHashtable::PtrList* list = get_entry(jt);
+  if (list != nullptr) {
+    // Add to existing list from the hash table:
+    list->add(om);
+    _om_count++;
+  } else {
+    // Create new list, add the ObjectMonitor* to it, and add the list to the hash table:
+    list = new (ResourceObj::C_HEAP, mtThread) ObjectMonitorsHashtable::PtrList();
+    list->add(om);
+    _om_count++;
+    add_entry(jt, list);
+  }
+}
+
+bool ObjectMonitorsHashtable::has_entry(JavaThread* jt, ObjectMonitor* om) {
+  ObjectMonitorsHashtable::PtrList* list = get_entry(jt);
+  if (list == nullptr || list->find(om) == nullptr) {
+    return false;
+  }
+  return true;
+}
+
 void MonitorList::add(ObjectMonitor* m) {
   ObjectMonitor* head;
   do {
@@ -992,6 +1030,8 @@ JavaThread* ObjectSynchronizer::get_lock_owner(ThreadsList * t_list, Handle h_ob
 
 // Visitors ...
 
+// This version of monitors_iterate() works with the in-use monitor list.
+//
 void ObjectSynchronizer::monitors_iterate(MonitorClosure* closure, JavaThread* thread) {
   MonitorList::Iterator iter = _in_use_list.iterator();
   while (iter.has_next()) {
@@ -999,6 +1039,30 @@ void ObjectSynchronizer::monitors_iterate(MonitorClosure* closure, JavaThread* t
     if (mid->owner() != thread) {
       continue;
     }
+    if (!mid->is_being_async_deflated() && mid->object_peek() != NULL) {
+      // Only process with closure if the object is set.
+
+      // monitors_iterate() is only called at a safepoint or when the
+      // target thread is suspended or when the target thread is
+      // operating on itself. The current closures in use today are
+      // only interested in an owned ObjectMonitor and ownership
+      // cannot be dropped under the calling contexts so the
+      // ObjectMonitor cannot be async deflated.
+      closure->do_monitor(mid);
+    }
+  }
+}
+
+// This version of monitors_iterate() works with the specified linked list.
+//
+void ObjectSynchronizer::monitors_iterate(MonitorClosure* closure,
+                                          ObjectMonitorsHashtable::PtrList* list,
+                                          JavaThread* thread) {
+  typedef LinkedListIterator<ObjectMonitor*> ObjectMonitorIterator;
+  ObjectMonitorIterator iter = ObjectMonitorIterator(list->head());
+  while (!iter.is_empty()) {
+    ObjectMonitor* mid = *iter.next();
+    assert(mid->owner() == thread, "must be");
     if (!mid->is_being_async_deflated() && mid->object_peek() != NULL) {
       // Only process with closure if the object is set.
 
@@ -1339,7 +1403,8 @@ void ObjectSynchronizer::chk_for_block_req(JavaThread* current, const char* op_n
 // Walk the in-use list and deflate (at most MonitorDeflationMax) idle
 // ObjectMonitors. Returns the number of deflated ObjectMonitors.
 size_t ObjectSynchronizer::deflate_monitor_list(Thread* current, LogStream* ls,
-                                                elapsedTimer* timer_p) {
+                                                elapsedTimer* timer_p,
+                                                ObjectMonitorsHashtable* table) {
   MonitorList::Iterator iter = _in_use_list.iterator();
   size_t deflated_count = 0;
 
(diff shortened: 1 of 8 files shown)


dcubed-ojdk:
I have logging in place for all of the RunThese configs to gather a little
bit of 'monitorinflation' info. So far a new sighting has not happened,
but I'm hopeful that the stars will align sometime soon.

I have learned thru my own test jobs that a RunThese24H job doesn't
always include calls to VM_ThreadDump::doit() so it seems that there's
yet another factor in these RunThese24H failures.

dcubed-ojdk:
The proposed solution in this patch is to call deflate_idle_monitors() from
VM_ThreadDump::doit() with a new and optional ObjectMonitorsHashtable
parameter. As deflate_idle_monitors() and deflate_monitor_list() do their
work, owned ObjectMonitors are added to the ObjectMonitorsHashtable
indexed by the owning JavaThread. After the deflate_idle_monitors() call
is finished VM_ThreadDump::doit() can find and return information about
the owned ObjectMonitors for each JavaThread without doing a linear walk
of the in-use monitor list for each JavaThread. A JavaThread* is used to
lookup a LinkedList in the ObjectMonitorsHashtable and if one exists, then
the JavaThread* owns all the ObjectMonitors on that LinkedList.

Also, the usual work done by deflate_idle_monitors() and deflate_monitor_list()
will serve to reduce the length of the in-use monitor list which makes the
in-use monitor list more efficient for other parts of the system.

dcubed-ojdk:
This fix has been tested with Mach5 Tier[1-8]. I've also done some targeted
memory leak testing since ObjectMonitorsHashtable is new (see the bug
report for details). I've rebased the fix to the latest JDK18 repo as of 2021.12.14
mid-day and I'm doing some sanity checking with Mach5 Tier[1-3].

It would be good to hear from @dholmes-ora, @coleenp, @fisk and @robehn
since they were my reviewers on:

JDK-8253064 monitor list simplifications and getting rid of TSM
https://bugs.openjdk.java.net/browse/JDK-8253064

Update: Mach5 Tier[1-3] testing looks good.

coleenp:
This looks good to me.  You solved both the problem of the very long monitor list and the per-thread iteration too. I like it!

dcubed-ojdk:
@coleenp - Thanks for the draft review! It's now officially ready...

src/hotspot/share/runtime/synchronizer.cpp
```diff
@@ -57,6 +57,44 @@
 #include "utilities/events.hpp"
 #include "utilities/preserveException.hpp"
 
+class CleanupObjectMonitorsHashtable: StackObj {
+ public:
+  bool do_entry(JavaThread*& key, ObjectMonitorsHashtable::PtrList*& list) {
```
dholmes-ora: I don't see this being called by anything ???

dcubed-ojdk: See L71 below:   _ptrs->unlink(&cleanup);  // cleanup the LinkedLists
That passes the cleanup object to ResourceHashtable::unlink()
function which calls the do_entry() function in the cleanup object.

dcubed-ojdk:
@dholmes-ora - Thanks for reviewing!

src/hotspot/share/runtime/synchronizer.cpp
dholmes-ora: Ah I see (well I don't really see as this is very obscure). I would have expected `CleanupObjectMonitorsHashtable` to be a subtype of something that I could then see is used by the ResourceHashtable. I'm going to guess that there is some template usage here that ensures the passed in object has the right API even if there is no subtyping relationship. Hard to see these connections at a distance.

To be clear not a review as I don't understand our Hashtable's enough to make much sense of what is actually happening here.

src/hotspot/share/runtime/vmOperations.cpp
```diff
@@ -279,6 +279,18 @@ void VM_ThreadDump::doit() {
     concurrent_locks.dump_at_safepoint();
   }
 
+  ObjectMonitorsHashtable table;
+  ObjectMonitorsHashtable* tablep = nullptr;
```
robehn, line 283: It looks like you can remove this pointer.
ThreadStackTrace::dump_stack_at_safepoint also looks at "with_locked_monitors", if false it ignores the table.
So there should not be any need to pass in null.

src/hotspot/share/services/threadService.cpp
```diff
@@ -683,9 +683,17 @@ void ThreadStackTrace::dump_stack_at_safepoint(int maxDepth) {
...
     // Iterate inflated monitors and find monitors locked by this thread
-    // not found in the stack
+    // and not found in the stack:
     InflatedMonitorsClosure imc(this);
-    ObjectSynchronizer::monitors_iterate(&imc, _thread);
+    if (table != nullptr) {
+      // Get the ObjectMonitors locked by this thread (if any):
+      ObjectMonitorsHashtable::PtrList* list = table->get_entry(_thread);
+      if (list != nullptr) {
+        ObjectSynchronizer::monitors_iterate(&imc, list, _thread);
```
robehn, line 694: The InflatedMonitorsClosure walks the stack until object is found with this method:
ThreadStackTrace::is_owned_monitor_on_stack(oop object)
for every object...

If you instead just collect all locks on stack with one pass you don't have to walk the stack over and over, which should be major speedup.

robehn (approved):

src/hotspot/share/runtime/synchronizer.cpp
coleenp: It would be a nice place for a lambda (on the list of future RFEs).  You could also put this class in scope of the function that calls it which might help.  (optional)

coleenp (approved):
This looks good.

src/hotspot/share/runtime/synchronizer.cpp
dcubed-ojdk: ResourceHashtable::unlink() is definitely one of those template
things that I don't really like, but I have to use it... All that stuff
is in src/hotspot/share/utilities/resourceHash.hpp, but I don't
see any precedent for putting the CleanupObjectMonitorsHashtable
definition in there. I think I'll leave it where it is.

src/hotspot/share/runtime/vmOperations.cpp
dcubed-ojdk: Right now `ThreadStackTrace::dump_stack_at_safepoint()` is flexible in
that it does not require `ObjectMonitorsHashtable* table` to be non-null
when `_with_locked_monitors` is true. 

From this call site, when `_with_locked_monitors` is true we always pass
a non-nullptr `ObjectMonitorsHashtable* table` and we do that because
we're at a safepoint and likely have to deal with a lot of threads if not all
the threads in the system.

For other `ThreadStackTrace::dump_stack_at_safepoint()` call sites,
I want things to be more flexible. Mostly because I haven't done
detailed analysis of them to justify changing their behavior.

src/hotspot/share/services/threadService.cpp
dcubed-ojdk: Sounds like an interesting RFE, but I'd prefer that be investigated
separately from this bug.

dcubed-ojdk:
@coleenp - Thanks for chiming in on the official review.

@robehn - Thanks for reviewing!

dcubed-ojdk:
Mach5 Tier[4567] testing looks good.
Mach5 Tier8 is still running and won't finish for at least another day or two.

Update: Mach5 Tier8 testing for v0 finished and looks good.

src/hotspot/share/services/threadService.cpp
dcubed-ojdk: Clarification: ThreadStackTrace::is_owned_monitor_on_stack() is not actually
walking the (JavaThread's) stack, but it is walking the collection of
StackFrameInfo objects that was gathered earlier in
ThreadStackTrace::dump_stack_at_safepoint(). Adding a gather-locked-monitor
helper object to the "add_stack_frame()" call would provide an easy hook and
then InflatedMonitorsClosure could use the new helper object. Something like
that anyway...

src/hotspot/share/runtime/synchronizer.cpp
```diff
@@ -57,6 +57,44 @@
...
+  CleanupObjectMonitorsHashtable cleanup;
+  _ptrs->unlink(&cleanup);  // cleanup the LinkedLists
+  delete _ptrs;             // then delete the hash table
+}
+
+void ObjectMonitorsHashtable::add_entry(JavaThread* jt, ObjectMonitor* om) {
+  ObjectMonitorsHashtable::PtrList* list = get_entry(jt);
+  if (list != nullptr) {
+    // Add to existing list from the hash table:
+    list->add(om);
```
fisk: These two lines are executed in both the if and else clause. Consider putting it after instead.

src/hotspot/share/runtime/synchronizer.cpp
```diff
@@ -1013,6 +1053,30 @@ void ObjectSynchronizer::monitors_iterate(MonitorClosure* closure, JavaThread* t
   }
 }
 
+// This version of monitors_iterate() works with the specified linked list.
+//
+void ObjectSynchronizer::monitors_iterate(MonitorClosure* closure,
+                                          ObjectMonitorsHashtable::PtrList* list,
+                                          JavaThread* thread) {
+  typedef LinkedListIterator<ObjectMonitor*> ObjectMonitorIterator;
+  ObjectMonitorIterator iter = ObjectMonitorIterator(list->head());
```
fisk: This seems to create an object and then copy it to iter. I think the more obvious syntax is ObjectMonitorIterator iter(list->head());

src/hotspot/share/runtime/synchronizer.cpp
```diff
@@ -1350,6 +1415,15 @@ size_t ObjectSynchronizer::deflate_monitor_list(Thread* current, LogStream* ls,
     ObjectMonitor* mid = iter.next();
     if (mid->deflate_monitor()) {
       deflated_count++;
+    } else if (table != nullptr) {
+      // The caller is interested in the owned ObjectMonitors. This does not capture
+      // unowned ObjectMonitors that cannot be deflated because of a waiter. Since
+      // deflate_idle_monitors() and deflate_monitor_list() can be called more than
+      // once, we have to make sure the entry has not already been added.
+      JavaThread* jt = (JavaThread*)mid->owner();
```
fisk: Hmmmm. I'm not sure this is correct. The owned of an ObjectMonitor is a JavaThread most of the time, but not always. I think that if the owner started off with a stack lock, and another thread inflated the lock due to contention, then the ObjectMonitor will get the stack lock as the owner, as the random other thread is too lazy to look up which thread stack the stack lock belongs to.

Because of that, I think we might miss locks that really are owned by a thread, as the thread key now will not map to such locks. Some kind of ownership lookup probably needs to happen here.

fisk (changes requested):
Looks okay, but I found a potential issue.

src/hotspot/share/services/threadService.cpp
robehn: Yes, sorry, we loop the stack frame info for every object. I.e. a loop in a loop which  is bad for time-complexity.
If the we created a sorted list from the frame info, we can binary search it instead.
Since a thread often only have at max a few locks, but could have hundreds of frames, just to be able skip looping over all frames without locks would be very beneficial.

src/hotspot/share/runtime/synchronizer.cpp
dcubed-ojdk: I'll take a look at doing that. Thanks for the suggestion.

src/hotspot/share/runtime/synchronizer.cpp
dcubed-ojdk: Nice catch. I'll fix that.

src/hotspot/share/runtime/synchronizer.cpp
dcubed-ojdk: Nice catch on an oddity of ObjectMonitor ownership!

The baseline code has the same "issue":

`void ObjectSynchronizer::monitors_iterate(MonitorClosure* closure, JavaThread* thread) {`

`<snip>`

```
    ObjectMonitor* mid = iter.next();
    if (mid->owner() != thread) {
      continue;
    }
```

Please notice that I said "issue" and not "bug". This isn't a bug and you really
won't like the reason that it's not a bug. I know that I don't like it...

Because of this ownership check `monitors_iterate()` won't apply the closure
(InflatedMonitorsClosure in this case) to an ObjectMonitor that is marked as
owned by a stack lock address because it will never match the target JavaThread*.

Here's the closure:
```
// Iterate through monitor cache to find JNI locked monitors
class InflatedMonitorsClosure: public MonitorClosure {
private:
  ThreadStackTrace* _stack_trace;
public:
  InflatedMonitorsClosure(ThreadStackTrace* st) {
    _stack_trace = st;
  }
  void do_monitor(ObjectMonitor* mid) {
    oop object = mid->object();
    if (!_stack_trace->is_owned_monitor_on_stack(object)) {
      _stack_trace->add_jni_locked_monitor(object);
    }
  }
};
```

This particular monitors_iterate() call uses InflatedMonitorsClosure to
gather just the JNI locked monitors owned by the target JavaThread*.
A JNI locked monitor will never have a stack lock address as its owner
marking. It will always have a JavaThread*. So the baseline code isn't
"broken" because it will find the JNI locked monitors just fine. Similarly
the new code isn't "broken" because it will also find the JNI locked
monitors just fine.

I definitely need to clarify the baseline comments and the new code
comments to make this subtlety much more clear.

dcubed-ojdk:
@fisk - Thanks for the review!

dcubed-ojdk:
I made changes to address @fisk's review. I've also included comment
updates to make clear the limitations of monitors_iterate(). I need a
re-review from @fisk and a re-review from @coleenp would also be
a nice thing to have. Thanks, in advance, for the re-review(s)...

Tested this latest patch with Mach5 Tier1. Mach5 Tier[23] are running now.

Update: Mach5 Tier[1-5] look good.

dcubed-ojdk:
The `ObjectMonitor::_owner` field can contain a `JavaThread*`, the address of
a `BasicLock` on a JavaThread's stack, `nullptr` or `DEFLATER_MARKER` so the
`_owner` field is of type `void*`.

@coleenp pointed out off-line that I had a bit of type confusion in my new
`ObjectMonitorsHashtable`. Instead of indexing the Hashtable with a
`JavaThread*`, I should be indexing the Hashtable with a `void*` to reflect
the fact that my `_owner` field can be a `JavaThread*` or the address of a
`BasicLock` on the JavaThread's stack.

My most recent push to this PR cleans up the type confusion. This update
is tested with Mach5 Tier1 as a sanity check.

Update: Also ran Mach5 Tier6 and Tier7 on this latest patch and those results
look good. Tier8 is running now.

Update: Tier8 results also look good.

fisk:
> I made changes to address @fisk's review. I've also included comment
> 
> updates to make clear the limitations of monitors_iterate(). I need a
> 
> re-review from @fisk and a re-review from @coleenp would also be
> 
> a nice thing to have. Thanks, in advance, for the re-review(s)...
> 
> 
> 
> Tested this latest patch with Mach5 Tier1. Mach5 Tier[23] are running now.
> 
> 
> 
> Update: Mach5 Tier[1-5] look good.

When the owner field of an ObjectMonitor is the address of a stack lock instead of a thread, it doesn't mean that it is a stack lock. It means that it *was* a stack lock when it was acquired by that thread, but has since then been upgraded to a full ObjectMonitor (by another thread), due to contention. But the fact that it *was* a stack lock is irrelevant in this context; we care about what it is now. And it is an inflated lock that just happens to have a slightly obscured owner field, as the owner thread was not easily available when it was inflated.

In other words, I still believe this is a bug. Such a monitor that gets upgraded from a stack lock to a full ObjectMonitor, can - unline actual stack locks - be involved in a deadlock that we are trying to find. If we don't associate such ObjectMonitor objects with its owner thread, then we will not be able to find the deadlock as we are supposed to with this code.

coleenp (approved):
Approve again. This makes me a lot happier.

coleenp:
Re: both Robbin's and Erik's comments. I think Robbin's should be an RFE and Erik's should be another bug report and fixed separately.  This change fixes the timeout that's been observed multiple times and preserves existing behavior (with existing bugs). I think some redesign is in order to make this correct all the time.

fisk:
> Re: both Robbin's and Erik's comments. I think Robbin's should be an RFE and Erik's should be another bug report and fixed separately. This change fixes the timeout that's been observed multiple times and preserves existing behavior (with existing bugs). I think some redesign is in order to make this correct all the time.

I'm okay with fixing the other bug in a separate RFR.

fisk (approved):
If the bug I pointed out is filed separately, then this looks good.

robehn (approved):

fisk:
For the record, Threads::owning_thread_from_monitor_owner seems to be the function to look up who *really* own an ObjectMonitor, based on the value in the owner field, to deal with the stuff I talked about.

dcubed-ojdk:
@coleenp, @fisk and @robehn - Thanks for the re-reviews and the approvals.

dcubed-ojdk:
I need to address these comments from @fisk:

> When the owner field of an ObjectMonitor is the address of a stack lock instead
> of a thread, it doesn't mean that it is a stack lock. It means that it was a stack
> lock when it was acquired by that thread, but has since then been upgraded to
> a full ObjectMonitor (by another thread), due to contention. But the fact that it
> was a stack lock is irrelevant in this context; we care about what it is now. And
> it is an inflated lock that just happens to have a slightly obscured owner field,
> as the owner thread was not easily available when it was inflated.

What @fisk says here is correct and I'm probably guilty of over simplifying what
I was trying to explain earlier. Here's what I said earlier for context:

>> The baseline code has the same "issue":
>>
>> void ObjectSynchronizer::monitors_iterate(MonitorClosure* closure, JavaThread* thread) {
>>
>> <snip>
>>
>>     ObjectMonitor* mid = iter.next();
>>     if (mid->owner() != thread) {
>>       continue;
>>     }
>> Please notice that I said "issue" and not "bug". This isn't a bug and you really
>> won't like the reason that it's not a bug. I know that I don't like it...
>> 
>> Because of this ownership check monitors_iterate() won't apply the closure
>> (InflatedMonitorsClosure in this case) to an ObjectMonitor that is marked as
>> owned by a stack lock address because it will never match the target JavaThread*.

@fisk, please notice that I don't say that this is a stack lock. I say this is

>> an ObjectMonitor that is marked as owned by a stack lock address

and that is the "bug" in your opinion. The baseline code and new code
do not recognize that the owner is not a JavaThread* and don't apply
the closure even though the stack lock address might belong to the
current target JavaThread.

I think we are in strong agreement about what the attributes of the "bug" are.


> In other words, I still believe this is a bug. Such a monitor that gets upgraded
> from a stack lock to a full ObjectMonitor, can - unline actual stack locks - be
> involved in a deadlock that we are trying to find. If we don't associate such
> ObjectMonitor objects with its owner thread, then we will not be able to find
> the deadlock as we are supposed to with this code.

You are missing part of the algorithm in the code that calls monitors_iterate().
The monitors_iterate() part ONLY EXISTS to find the JNI locked ObjectMonitors.
It does that by finding all of the ObjectMonitors on the in-use list that are owned
by the target JavaThread* (and by owned I mean _owner == JavaThread*). It uses
the closure to search for each matched ObjectMonitor* in the JavaThread's stack
info that it squirreled away before making the call to monitors_iterate(). If it finds
the ObjectMonitor* on the JavaThread's stack, then that ObjectMonitor* IS NOT
a JNI locked ObjectMonitor and monitors_iterate() ignores it.

If monitors_iterate() does not find the ObjectMonitor* in the JavaThread's stack
info, then it assumes that the ObjectMonitor* is a JNI locked monitor and it adds
that ObjectMonitor* to the thread dump info.

Because an ObjectMonitor that is marked as owned by a stack lock address won't
match the target JavaThread*, it won't get processed by the closure that is called
by monitors_iterate(). However, we don't care because that ObjectMonitor* IS NOT
a JNI locked monitor and that is what monitors_iterate() is looking for. It must be a
monitor that is mentioned in the JavaThread's stack because it was acquired that
way originally so it will ALREADY be included in the thread's info that was gathered
before we made the monitors_iterate() call. Nothing is missed here!

I have tried to clarify this in the comments that I added as part of this patch. I'll
re-review the comments again and see if there is anything else I can add or change.

dcubed-ojdk:
I've filed the following RFE for @robehn's improvement:

JDK-8278968 InflatedMonitorsClosure repeats walks of StackFrameInfo and should cache instead
https://bugs.openjdk.java.net/browse/JDK-8278968

fisk (approved):
I guess you can recursively JNI lock an ObjectMonitor, that has been promoted from a stack lock. Then it's a lock held by JNI and we won't report it here.

But this is JVMTI code so there is probably some other random check somewhere else that checks for that so I'm just going to assume this works somehow.

Ship it!

dcubed-ojdk:
If you recursively JNI lock an ObjectMonitor that has been promoted from a stack lock,
then I think the code will update the _owner field from the address of the original stack
lock to the owning JavaThread*. Only the owning JavaThread can recursively enter an
existing ObjectMonitor and only the owning JavaThread can update that _owner field...

dcubed-ojdk:
@fisk - Thanks for closing the loop on this thread.

PR #27, opened 2021-12-15
8277964: ClassCastException with no stack trace is thrown with -Xcomp in method handle invocation

vnkozlov:
A proper fix for this is to use the catchException combination. However, that introduces significant cold startup performance regression. JDK-8278447 tracks the work to address the performance regression using catchException and asSpreader combinator. It may require significant work and refactoring which is risky for JDK 18. 

It is proposed to implement a workaround in C2 to white list the relevant methods (all methods in sun.invoke.util.ValueConversions class) not to omit stack trace when exception is thrown in them.

Added new regression test. Tested tier1-3.

---------
### Progress

### Issue
 * [JDK-8277964](https://bugs.openjdk.java.net/browse/JDK-8277964): ClassCastException with no stack trace is thrown with -Xcomp in method handle invocation

### Reviewers
 * [Dean Long](https://openjdk.java.net/census#dlong) (@dean-long - **Reviewer**) ⚠️ Review applies to 3c23350d358a0973ddb4d1ea9b6571dd597ba5b5
 * [Mandy Chung](https://openjdk.java.net/census#mchung) (@mlchung - **Reviewer**)
 * [David Holmes](https://openjdk.java.net/census#dholmes) (@dholmes-ora - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/27/head:pull/27` \
`$ git checkout pull/27`

`$ git checkout pull/27` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/27/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 27`

`$ git pr show -t 27`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/27.diff">https://git.openjdk.java.net/jdk18/pull/27.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/oops/method.cpp
+++ b/src/hotspot/share/oops/method.cpp
@@ -824,10 +824,8 @@ bool Method::can_be_statically_bound(InstanceKlass* context) const {
  *  Returns true otherwise.
  */
 bool Method::can_omit_stack_trace() {
-  if (method_holder()->class_loader_data()->is_boot_class_loader_data()) {
-    if (klass_name() == vmSymbols::sun_invoke_util_ValueConversions()) {
-      return false; // All methods in sun.invoke.util.ValueConversions
-    }
+  if (klass_name() == vmSymbols::sun_invoke_util_ValueConversions()) {
+    return false; // All methods in sun.invoke.util.ValueConversions
   }
   return true;
 }
--- a/src/hotspot/share/oops/method.cpp
+++ b/src/hotspot/share/oops/method.cpp
@@ -818,6 +818,20 @@ bool Method::can_be_statically_bound(InstanceKlass* context) const {
   return (method_holder() == context) && can_be_statically_bound();
 }
 
+/**
+ *  Returns false if this is one of specially treated methods for
+ *  which we have to provide stack trace in throw in compiled code.
+ *  Returns true otherwise.
+ */
+bool Method::can_omit_stack_trace() {
+  if (method_holder()->class_loader_data()->is_boot_class_loader_data()) {
+    if (klass_name() == vmSymbols::sun_invoke_util_ValueConversions()) {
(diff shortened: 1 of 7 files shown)


src/hotspot/share/oops/method.cpp
```diff
@@ -818,6 +818,20 @@ bool Method::can_be_statically_bound(InstanceKlass* context) const {
   return (method_holder() == context) && can_be_statically_bound();
 }
 
+/**
+ *  Returns false if this is one of specially treated methods for
+ *  which we have to provide stack trace in throw in compiled code.
+ *  Returns true otherwise.
+ */
+bool Method::can_omit_stack_trace() {
+  if (method_holder()->class_loader_data()->is_boot_class_loader_data()) {
```
dholmes-ora: Do you actually need to check this?

dean-long (approved):

mlchung (approved):
Looks good.   Thanks for taking this on.

src/hotspot/share/oops/method.cpp
vnkozlov: @dholmes-ora, thank you for looking on it.
I discussed it with Mandy and agreed that we need to narrow down this workaround as much as possible. That is why it is done only for system class loaded by null loader.

mlchung: David has a good observation.   There will be no split package for modules.  So sun.invoke.util classes will only be loaded from java.base.  The boot loader is not needed.

vnkozlov: Okay, I will remove it.

mlchung (approved):

dholmes-ora (approved):
Looks good.

Thanks,
David

vnkozlov:
Thank you, David, Dean and Mandy for reviews.

PR #29, opened 2021-12-15
8275638: GraphKit::combine_exception_states fails with "matching stack sizes" assert

rwestrel:
The bug and fix were discussed in a previous PR:

https://github.com/openjdk/jdk/pull/6572

I pushed all commits from that PR on top of jdk 18 and added a couple
extra tests as suggested in:

https://github.com/openjdk/jdk/pull/6572#issuecomment-994086590

---------
### Progress

### Issue
 * [JDK-8275638](https://bugs.openjdk.java.net/browse/JDK-8275638): GraphKit::combine_exception_states fails with "matching stack sizes" assert

### Reviewers
 * [Dean Long](https://openjdk.java.net/census#dlong) (@dean-long - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/29/head:pull/29` \
`$ git checkout pull/29`

`$ git checkout pull/29` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/29/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 29`

`$ git pr show -t 29`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/29.diff">https://git.openjdk.java.net/jdk18/pull/29.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -416,6 +416,7 @@ class LateInlineMHCallGenerator : public LateInlineCallGenerator {
 };
 
 bool LateInlineMHCallGenerator::do_late_inline_check(Compile* C, JVMState* jvms) {
+  assert(!jvms->method()->has_exception_handlers(), "");
   // Even if inlining is not allowed, a virtual call can be strength-reduced to a direct call.
   bool allow_inline = C->inlining_incrementally();
   bool input_not_const = true;
--- a/src/hotspot/share/opto/callGenerator.cpp
+++ b/src/hotspot/share/opto/callGenerator.cpp
@@ -416,7 +416,11 @@ class LateInlineMHCallGenerator : public LateInlineCallGenerator {
 };
 
 bool LateInlineMHCallGenerator::do_late_inline_check(Compile* C, JVMState* jvms) {
-  assert(!jvms->method()->has_exception_handlers(), "");
+  // When inlining a virtual call, the null check at the call and the call itself can throw. These 2 paths have different
+  // expression stacks which causes late inlining to break. The MH invoker is not expected to be called from a method wih
+  // exception handlers. When there is no exception handler, GraphKit::builtin_throw() pops the stack which solves the issue
+  // of late inlining with exceptions.
+  assert(!jvms->method()->has_exception_handlers(), "no exception handler expected");
   // Even if inlining is not allowed, a virtual call can be strength-reduced to a direct call.
   bool allow_inline = C->inlining_incrementally();
   bool input_not_const = true;
--- a/src/hotspot/share/opto/graphKit.cpp
+++ b/src/hotspot/share/opto/graphKit.cpp
@@ -345,7 +345,6 @@ void GraphKit::combine_exception_states(SafePointNode* ex_map, SafePointNode* ph
   JVMState* ex_jvms = ex_map->_jvms;
   assert(ex_jvms->same_calls_as(phi_map->_jvms), "consistent call chains");
   assert(ex_jvms->stkoff() == phi_map->_jvms->stkoff(), "matching locals");
-//  assert(ex_jvms->sp() == phi_map->_jvms->sp(), "matching stack sizes");
   assert(ex_jvms->monoff() == phi_map->_jvms->monoff(), "matching JVMS");
   assert(ex_jvms->scloff() == phi_map->_jvms->scloff(), "matching scalar replaced objects");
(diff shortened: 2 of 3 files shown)


dean-long (approved):

vnkozlov (approved):
I agree with this conservative fix. John agreed with it too.
@dean-long please run testing for it before Roland pushed it. Also file RFE to clean these up - we will assign an engineer to look on it later.

dean-long:
I filed JDK-8278873 and started testing.

dean-long:
Test results look good.

rwestrel:
@dean-long @vnkozlov thanks for the reviews.

iwanowww:
I'm late to the party, but still would like to clarify one thing.

It seems the root cause of the bug comes from the fact that the same JVM state is used by both `GraphKit::uncommon_trap()` and `GraphKit::builtin_throw()` while `GraphKit::null_check_receiver_before_call()` deliberately adjusts the state to please the former case. If the original state (after the call) is used for `GraphKit::builtin_throw()`, it should fix the bug as well, shouldn't it?

```
  // Do a null check on the receiver as it would happen before the call to
  // callee (with all arguments still on the stack).
  Node* null_check_receiver_before_call(ciMethod* callee) {
    assert(!callee->is_static(), "must be a virtual method");
    // Callsite signature can be different from actual method being called (i.e _linkTo* sites).
    // Use callsite signature always.
    ciMethod* declared_method = method()->get_method_at_bci(bci());
    const int nargs = declared_method->arg_size();
    inc_sp(nargs);
    Node* n = null_check_receiver();
    dec_sp(nargs);
    return n;
  }
```

rwestrel:
> It seems the root cause of the bug comes from the fact that the same JVM state is used by both `GraphKit::uncommon_trap()` and `GraphKit::builtin_throw()` while `GraphKit::null_check_receiver_before_call()` deliberately adjusts the state to please the former case. If the original state (after the call) is used for `GraphKit::builtin_throw()`, it should fix the bug as well, shouldn't it?

Wouldn't that only work if there's no uncommon trap in GraphKit::builtin_throw()?

iwanowww:
> Wouldn't that only work if there's no uncommon trap in GraphKit::builtin_throw()?

Good point. I overlooked the uncommon trap logic there. 

But my original point still stands: why can't `GraphKit::builtin_throw()` be taught to distinguish between throwing and uncommon trap cases and adjust JVM state accordingly? In this particular case, it seems simpler to start with the JVM state for re-execution (before call) and switch to the JVM state after the call (pop arguments from the stack) when an exception is thrown.

So, instead of checking for local exception handlers and wiping the stack [1], the effects of `null_check_receiver_before_call()` [2] can be undone unconditionally. That should make all possible exception states uniform and, hence, mergeable, shouldn't it?    

[1]
```
       if (!method()->has_exception_handlers()) {
        // We don't need to preserve the stack if there's no handler as the entire frame is going to be popped anyway.
        // This prevents issues with exception handling and late inlining.
        set_sp(0);
        clean_stack(0);
      }
``` 

[2] 
```
  // Do a null check on the receiver as it would happen before the call to
  // callee (with all arguments still on the stack).
  Node* null_check_receiver_before_call(ciMethod* callee) {
    assert(!callee->is_static(), "must be a virtual method");
    // Callsite signature can be different from actual method being called (i.e _linkTo* sites).
    // Use callsite signature always.
    ciMethod* declared_method = method()->get_method_at_bci(bci());
    const int nargs = declared_method->arg_size();
    inc_sp(nargs);
    Node* n = null_check_receiver();
    dec_sp(nargs);
    return n;
  }
```

rwestrel:
> But my original point still stands: why can't `GraphKit::builtin_throw()` be taught to distinguish between throwing and uncommon trap cases and adjust JVM state accordingly? In this particular case, it seems simpler to start with the JVM state for re-execution (before call) and switch to the JVM state after the call (pop arguments from the stack) when an exception is thrown.

Then what's on the stack doesn't matter so why not pop the stack instead?
There's some discussion about that in the initial PR (before the jdk18 fork): https://github.com/openjdk/jdk/pull/6572
And a follow up bug: https://bugs.openjdk.java.net/browse/JDK-8278846

iwanowww:
> Then what's on the stack doesn't matter so why not pop the stack instead?

Yes, that's also an option. (And, probably, the cleanest one.) 

I agree with the summary in JDK-8278846. I don't see any reason why exception state should include stack state. 

Thanks for the clarifications.

PR #30, opened 2021-12-15
8278413: C2 crash when allocating array of size too large

rwestrel:
On the fallthrough path from an AllocateArray, the length of the
allocated array is casted (with a CastII) to [0, max_size] with
max_size some number that depends on the array type and can be less
than max_jint.

Allocating an array of a length that's not in [0, max_size] causes the
CastII to become top. The fallthrough path must be killed as well in
that case otherwise this can lead to a broken graph. Currently c2 has
logic to protect against an allocation of array of negative size in
AllocateArrayNode::Ideal(). That call replaces the fallthrough path
with an Halt node. But if the size is too big, then the fallthrough
path is left as is.

This patch fixes that issues. It also reworks the length negative
case. I added a Bool/CmpU input to the AllocateArray that tests for a
valid length. If that input becomes false, CatchNode::Value() kills
the fallthrough path. That logic is similar to that for a virtual call
with a null receiver. I also removed AllocateArrayNode::Ideal() now
that CatchNode::Value() takes care of the same corner case. The code
in AllocateArrayNode::Ideal() was added by Vladimir and he told me he
tried extending CatchNode::Value() at the time but that caused test
failures. I had no issues in my testing so I assume doing it that way
is ok now.

The new input to AllocateArray is moved to the CallStaticJava runtime
call for array allocation on macro expansion as a precedence edge. The
reason for that is that final graph reshape needs a way to tell
whether the missing path out of the allocation is legal or not. final
graph reshape then removes the then useless precedence edge.

---------
### Progress

### Issue
 * [JDK-8278413](https://bugs.openjdk.java.net/browse/JDK-8278413): C2 crash when allocating array of size too large

### Reviewers
 * [Nils Eliasson](https://openjdk.java.net/census#neliasso) (@neliasso - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/30/head:pull/30` \
`$ git checkout pull/30`

`$ git checkout pull/30` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/30/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 30`

`$ git pr show -t 30`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/30.diff">https://git.openjdk.java.net/jdk18/pull/30.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/opto/cfgnode.cpp
+++ b/src/hotspot/share/opto/cfgnode.cpp
@@ -2687,6 +2687,20 @@ const Type* CatchNode::Value(PhaseGVN* phase) const {
       // Rethrows always throw exceptions, never return
       if (call->entry_point() == OptoRuntime::rethrow_stub()) {
         f[CatchProjNode::fall_through_index] = Type::TOP;
+      } else if (call->is_AllocateArray()) {
+        Node* klass_node = call->in(AllocateNode::KlassNode);
+        Node *length = call->in(AllocateNode::ALength);
+        const Type* length_type = phase->type(length);
+        const Type* klass_type = phase->type(klass_node);
+        if (length_type == Type::TOP || klass_type == Type::TOP) {
+          f[CatchProjNode::fall_through_index] = Type::TOP;
+        } else {
+          Node* valid_length_test = call->in(AllocateNode::ValidLengthTest);
+          const Type* valid_length_test_t = phase->type(valid_length_test);
+          if (valid_length_test_t->isa_int() && valid_length_test_t->is_int()->is_con(0)) {
+            f[CatchProjNode::fall_through_index] = Type::TOP;
+          }
+        }
       } else if( call->req() > TypeFunc::Parms ) {
         const Type *arg0 = phase->type( call->in(TypeFunc::Parms) );
         // Check for null receiver to virtual or interface calls
--- a/src/hotspot/share/opto/callnode.cpp
+++ b/src/hotspot/share/opto/callnode.cpp
@@ -1656,6 +1656,7 @@ AllocateNode::AllocateNode(Compile* C, const TypeFunc *atype,
   init_req( KlassNode          , klass_node);
   init_req( InitialTest        , initial_test);
   init_req( ALength            , topnode);
+  init_req( ValidLengthTest    , topnode);
   C->add_macro_node(this);
 }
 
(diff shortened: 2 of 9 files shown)


neliasso:
Have you run any micros to verify that nothing regresses?

src/hotspot/share/opto/cfgnode.cpp
```diff
@@ -2687,6 +2687,20 @@ const Type* CatchNode::Value(PhaseGVN* phase) const {
...
+        Node* klass_node = call->in(AllocateNode::KlassNode);
+        Node *length = call->in(AllocateNode::ALength);
+        const Type* length_type = phase->type(length);
+        const Type* klass_type = phase->type(klass_node);
+        if (length_type == Type::TOP || klass_type == Type::TOP) {
+          f[CatchProjNode::fall_through_index] = Type::TOP;
+        } else {
+          Node* valid_length_test = call->in(AllocateNode::ValidLengthTest);
+          const Type* valid_length_test_t = phase->type(valid_length_test);
+          if (valid_length_test_t->isa_int() && valid_length_test_t->is_int()->is_con(0)) {
```
neliasso, line 2700: Here you do: 
```
Node* valid_length_test = call->in(AllocateNode::ValidLengthTest);
const Type* valid_length_test_t = phase->type(valid_length_test);
if (valid_length_test_t->isa_int() && valid_length_test_t->is_int()->is_con(0)) {
```

But in compile.cpp:3766 you do:
```
Node* valid_length_test = call->in(call->req());
call->rm_prec(call->req());
if (valid_length_test->find_int_con(1) == 0) {
```

Why "call->req()" and not "call->in(AllocateNode::ValidLengthTest)"?

And why not "if (valid_length_test->find_int_con(1) == 0) {" in both places?

neliasso (changes requested):
I general I really like the change. I simplifies things nicely.

rwestrel:
> Have you run any micros to verify that nothing regresses?

Thanks for looking at this. This fixes a crash for a corner case (array size close to max_jint). I don't think it can affect performance so I haven't run any performance testing.

src/hotspot/share/opto/cfgnode.cpp
rwestrel: > Here you do:
> 
> ```
> Node* valid_length_test = call->in(AllocateNode::ValidLengthTest);
> const Type* valid_length_test_t = phase->type(valid_length_test);
> if (valid_length_test_t->isa_int() && valid_length_test_t->is_int()->is_con(0)) {
> ```
> 
> But in compile.cpp:3766 you do:
> 
> ```
> Node* valid_length_test = call->in(call->req());
> call->rm_prec(call->req());
> if (valid_length_test->find_int_con(1) == 0) {
> ```
> 
> Why "call->req()" and not "call->in(AllocateNode::ValidLengthTest)"?

call->in(AllocateNode::ValidLengthTest) only works for allocateArrayNode. The code I modified in compile.cpp runs after macro expansion so there's no AllocateArrayNode anymore. Instead there's a call to the runtime. When the AllocateArrayNode is macro expanded I move in(AllocateNode::ValidLengthTest) to the new runtime call as a precedence edge that is at req(). I tried adding it as an extra parameter that would be removed in compile.cpp but that messed up debug infos and I hit some asserts.

> > And why not "if (valid_length_test->find_int_con(1) == 0) {" in both places?

So`that the Value() call does the right thing during CCP where a node can be registered as a constant but not transformed yet to a ConNode.

neliasso:
Passes testing tier1-3

src/hotspot/share/opto/cfgnode.cpp
neliasso: That's make sense. Thanks for explaining!

neliasso (approved):
Looks good!

vnkozlov (approved):
Looks good. Thank you for fixing it.

rwestrel:
@neliasso @vnkozlov thanks for the reviews.

PR #41, opened 2021-12-17
8278267: ARM32: several vector test failures for ASHR

shqking:
In ARM32, "VSHL (register)" instruction [1] is shared by vector left
shift and vector right shift, and the condition to distinguish them is
whether the shift count value is positve or negative. Hence, negation
operation is needed before conducting vector right shift.

For vector right shift, the shift count can be a RShiftCntV or a normal
vector node. Take test case Byte64VectorTests.java [2][3] as an example.
Note that RShiftCntV is already negated via rules "vsrcntD" and
"vsrcntX" whereas the normal vector node is NOT, since we don't know
whether a normal vector node is used as a vector shift count or not.
This is the root cause for these vector test failures.

The fix is simple, moving the negation from "vsrcntD|X" to the
corresponding vector right shift rules.

Affected rules are vsrlBB_reg and vsraBB_reg. Note that vector shift
related rules are in form of "vsAABB_CC", where
1) AA can be l (left shift), rl (logical right shift) and ra (arithmetic
  right shift).
2) BB can be 8B/16B (byte type), 4S/8S (short type), 2I/4I (int type)
  and 2L (long type).
3) CC can be reg (register case) and immI (immediate case).

Minor updates:
1) Merge "vslcntD" and "vsrcntD" into rule "vscntD", as these two rules
conduct the same duplication operation now.
2) Update the "match" primitive for vsraBB_immI rules.
3) Style issue: remove the surrounding space for "ins_pipe" primitive.

Tests:
We ran tier 1~3 tests on ARM32 platform. With this patch, previously
failed vector test cases can pass now without introducing test
regression.

[1] https://developer.arm.com/documentation/ddi0406/c/Application-Level-Architecture/Instruction-Details/Alphabetical-list-of-instructions/VSHL--register-?lang=en
[2] https://github.com/openjdk/jdk/blame/master/test/jdk/jdk/incubator/vector/Byte64VectorTests.java#L2237
[3] https://github.com/openjdk/jdk/blame/master/test/jdk/jdk/incubator/vector/Byte64VectorTests.java#L2425

---------
### Progress

### Issue
 * [JDK-8278267](https://bugs.openjdk.java.net/browse/JDK-8278267): ARM32: several vector test failures for ASHR

### Reviewers
 * [Ningsheng Jian](https://openjdk.java.net/census#njian) (@nsjian - Committer) ⚠️ Review applies to 3d29fb2ccf303942ecbd104ed8639983cb3d7a60
 * [Dean Long](https://openjdk.java.net/census#dlong) (@dean-long - **Reviewer**) ⚠️ Review applies to 566efefe2096237f86d0643e4d288e6aa49d1578

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/41/head:pull/41` \
`$ git checkout pull/41`

`$ git checkout pull/41` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/41/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 41`

`$ git pr show -t 41`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/41.diff">https://git.openjdk.java.net/jdk18/pull/41.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/arm/arm.ad
+++ b/src/hotspot/cpu/arm/arm.ad
@@ -10593,9 +10593,12 @@ instruct vneg16B_reg(vecX dst, vecX src) %{
 
 // ------------------------------ Shift ---------------------------------------
 
-instruct vslcntD(vecD dst, iRegI cnt) %{
+// Low bits of vector "shift" elements are used, so it
+// doesn't matter if we treat it as ints or bytes here.
+instruct vscntD(vecD dst, iRegI cnt) %{
   predicate(n->as_Vector()->length_in_bytes() == 8 && VM_Version::has_simd());
   match(Set dst (LShiftCntV cnt));
+  match(Set dst (RShiftCntV cnt));
   size(4);
   ins_cost(DEFAULT_COST); // FIXME
   expand %{
@@ -10603,9 +10606,10 @@ instruct vslcntD(vecD dst, iRegI cnt) %{
   %}
 %}
 
-instruct vslcntX(vecX dst, iRegI cnt) %{
+instruct vscntX(vecX dst, iRegI cnt) %{
   predicate(n->as_Vector()->length_in_bytes() == 16 && VM_Version::has_simd());
   match(Set dst (LShiftCntV cnt));
+  match(Set dst (RShiftCntV cnt));
   size(4);
   ins_cost(DEFAULT_COST); // FIXME
   expand %{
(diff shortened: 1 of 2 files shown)


chhagedorn:
@nick-arm could you or someone else at ARM help to review this? That would be great.

nsjian (approved):
The fix looks good to me (not a Reviewer).

dean-long:
There seems to be an interesting history here.

Method 1: negate in RShiftCntV rule
Method 2: negate in RShiftV* rules

For aarch64, the negate was moved into the shift instruction in JDK-8213134 (Method 1 --> Method 2).  Then JDK-8262916 proposed to move it back out of the shift instruction again.  In that PR, the opinion that Method 1 (arm32) generated better code than Method 2 (aarch64) was expressed.  Now it looks like this PR proposes for arm32 to move to Method 2 like aarch64, so I suspect that there will be a performance impact.

I think there is a simpler fix that doesn't require moving where the negate happens.  JDK-8277239 fixed a similar problem by adding a flag on vector shift nodes to indicate variable shift, then checking the flag in the predicate.  Perhaps arm32 could do the same?

nsjian:
> 
> I think there is a simpler fix that doesn't require moving where the negate happens. JDK-8277239 fixed a similar problem by adding a flag on vector shift nodes to indicate variable shift, then checking the flag in the predicate. Perhaps arm32 could do the same?

Hmm, I didn't notice JDK-8277239 before. That looks like a good solution. If it works fine on arm32, we could also do it on aarch64.

shqking:
> There seems to be an interesting history here.
> 
> Method 1: negate in RShiftCntV rule Method 2: negate in RShiftV* rules
> 
> For aarch64, the negate was moved into the shift instruction in JDK-8213134 (Method 1 --> Method 2). Then JDK-8262916 proposed to move it back out of the shift instruction again. In that PR, the opinion that Method 1 (arm32) generated better code than Method 2 (aarch64) was expressed. Now it looks like this PR proposes for arm32 to move to Method 2 like aarch64, so I suspect that there will be a performance impact.
> 
> I think there is a simpler fix that doesn't require moving where the negate happens. JDK-8277239 fixed a similar problem by adding a flag on vector shift nodes to indicate variable shift, then checking the flag in the predicate. Perhaps arm32 could do the same?

Thanks a lot for your explanation. I do agree with your solution.

Updated the code: 1) using is_var_shift() to determine to put "negation" in RShiftCntV or RShiftV* rules, and 2) adding `cmp` and `hash` for ShiftV node.

Could you please take a look at the latest code when you have a chance? Maybe after the new year :)
Thanks.

shqking:
Tests on the latest code: 
We ran tier 1~3 tests on linux+ARM32 platform. With this patch, previously failed vector test cases can pass now without introducing test regression.

bulasevich:
The change is Ok for me (I am not a reviewer).
Though I would avoid changing the style. If we want to change a style, I would suggest to change all the entries (there is a lot of surrounding spaces remaining in a file), and I think FIXME for ins_cost/ins_pipe are outdated as well.

shqking:
> The change is Ok for me (I am not a reviewer). Though I would avoid changing the style. If we want to change a style, I would suggest to change all the entries (there is a lot of surrounding spaces remaining in a file), and I think FIXME for ins_cost/ins_pipe are outdated as well.

Thanks for your review.
Yes. We may want to use another RFE to change the style issue for the whole file.
Hence, in the latest commit I made minimal updates to existing rules, and only updated the style in the newly added rules.

src/hotspot/cpu/arm/arm.ad
```diff
@@ -10790,7 +10792,7 @@ instruct vsl16B_reg(vecX dst, vecX src, vecX shift) %{
 %}
 
 instruct vsl8B_immI(vecD dst, vecD src, immI shift) %{
-  predicate(n->as_Vector()->length() == 8);
+  predicate(n->as_Vector()->length() == 8 && !n->as_ShiftV()->is_var_shift());
```
dean-long: Is checking for is_var_shift() really necessary for these immI rules?  It seems like it should never happen, so it could be an assert.

src/hotspot/share/opto/vectornode.hpp
```diff
@@ -531,16 +531,21 @@ class SqrtVDNode : public VectorNode {
...
   ShiftVNode(Node* in1, Node* in2, const TypeVect* vt, bool is_var_shift) :
     VectorNode(in1,in2,vt), _is_var_shift(is_var_shift) {
     init_class_id(Class_ShiftV);
   }
   virtual Node* Identity(PhaseGVN* phase);
   virtual int Opcode() const = 0;
+  virtual uint hash() const { return VectorNode::hash() + _is_var_shift; }
+  virtual bool cmp(const Node& n) const {
+    return VectorNode::cmp(n) && _is_var_shift == ((ShiftVNode&)n)._is_var_shift;
+  }
```
dean-long, line 546: Is it important to add hash() and cmp()?

src/hotspot/cpu/arm/arm.ad
```diff
@@ -10965,9 +10966,79 @@ instruct vsl2L_immI(vecX dst, vecX src, immI shift) %{
...
+instruct vsrl4S_reg_var(vecD dst, vecD src, vecD shift, vecD tmp) %{
+  predicate(n->as_Vector()->length() == 4 && n->as_ShiftV()->is_var_shift());
+  match(Set dst (URShiftVS src shift));
+  effect(TEMP tmp);
+  size(4*2);
+  ins_cost(DEFAULT_COST*2);
+  format %{
+    "VNEG.S8 $tmp.D,$shift.D\n\t! neg packed8B"
+    "VSHL.U16 $dst.D,$src.D,$tmp.D\t! logical right shift packed4S"
+  %}
```
dean-long, line 11008: This matches what aarch64 is doing, right?  I was hoping there was a way for multiple shift instructions to reuse the same negated value, like we do for vsrcntX.  But it would probably require introducing a new operand type, like vecNegatedRShiftX, or doing the negate in the IR like we do for the mask.

dean-long (approved):

src/hotspot/cpu/arm/arm.ad
shqking: An assertion should be better, but I didn't find a proper way to add one assertion for `instruct` according to `src/hotspot/share/adlc/Doc/Syntax.doc`.

Hence I'd like to put this check to `predicate`, and "bad AD file` assertion would be issued for unexpected cases.

src/hotspot/share/opto/vectornode.hpp
shqking: I think it's safe to add them to avoid mis-optimizations at IR level.

Per my understanding, two ShiftV nodes with the same vector elements but with different `_is_var_shift` field values, are two different nodes, and they cannot be merged/treated as the same node.

src/hotspot/cpu/arm/arm.ad
dean-long: You can put the assert into a helper function.  You can look in aarch64.ad for examples.

bool assert_not_var_shift(const Node *n) {
  assert(!n->as_ShiftV()->is_var_shift(), "illegal var shift");
  return true;
}

[...]

predicate(n->as_Vector()->length() == 8 && assert_not_var_shift(n));

src/hotspot/share/opto/vectornode.hpp
dean-long: That makes sense.  Is it possible to write a test for that?

src/hotspot/cpu/arm/arm.ad
shqking: Yes. AArch64 works in the same way.

Agree with you. Performance benefit would be obtained if we can reuse the same negated value.

We may want to prepare another patch to enhance right shifts for aarch64(using is_var_shift() and considering your proposed two solutions).

src/hotspot/cpu/arm/arm.ad
shqking: Thanks for you suggestion. Will update soon.

src/hotspot/share/opto/vectornode.hpp
shqking: Generally speaking `hash()` and `cmp()` need be overrided if there are new field members. E.g., LoadLNode, VectorMaskCmpNode, VectorTestNode, etc. 

It could happen theoretically. But I'm afraid it's not easy to design a test case as it might involve some internal optimizations in a crafty way to construct such two kinds of **ShiftV** nodes.

src/hotspot/cpu/arm/arm.ad
dean-long: A separate RFE for that sounds good to me.  Please file one.

shqking: I'd like to reuse this open JBS issue. See https://bugs.openjdk.java.net/browse/JDK-8265263 . And I will work on it soon.

dean-long: Perfect.  I forgot about that RFE.

src/hotspot/cpu/arm/arm.ad
shqking: @dean-long I believe your comments are addressed in the latest commit. Could you please help to take a review when you got a chance? Thanks.

dean-long: Yes, your changes look good.  Approved.

shqking:
Thanks for your reviews! @dean-long @nsjian 
I suppose this patch is ready to be integrated.

/integrate

PR #45, opened 2021-12-17
JDK-8278967 rmiregistry fails to start because SecurityManager is disabled

stuart-marks:
Enable the security manager in rmiregistry's launcher arguments.

---------
### Progress

### Issue
 * [JDK-8278967](https://bugs.openjdk.java.net/browse/JDK-8278967): rmiregistry fails to start because SecurityManager is disabled

### Reviewers
 * [Alan Bateman](https://openjdk.java.net/census#alanb) (@AlanBateman - **Reviewer**)
 * [Erik Joelsson](https://openjdk.java.net/census#erikj) (@erikj79 - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/45/head:pull/45` \
`$ git checkout pull/45`

`$ git checkout pull/45` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/45/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 45`

`$ git pr show -t 45`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/45.diff">https://git.openjdk.java.net/jdk18/pull/45.diff</a>

</details>

### DIFF
--- a/make/modules/java.rmi/Launcher.gmk
+++ b/make/modules/java.rmi/Launcher.gmk
@@ -27,5 +27,5 @@ include LauncherCommon.gmk
 
 $(eval $(call SetupBuildLauncher, rmiregistry, \
     MAIN_CLASS := sun.rmi.registry.RegistryImpl, \
-    JAVA_ARGS := -Djava.security.manager, \
+    JAVA_ARGS := -Djava.security.manager=allow, \
 ))
--- a/make/modules/java.rmi/Launcher.gmk
+++ b/make/modules/java.rmi/Launcher.gmk
@@ -27,4 +27,5 @@ include LauncherCommon.gmk
 
 $(eval $(call SetupBuildLauncher, rmiregistry, \
     MAIN_CLASS := sun.rmi.registry.RegistryImpl, \
+    JAVA_ARGS := -Djava.security.manager, \
 ))
--- a/test/jdk/tools/launcher/VersionCheck.java
+++ b/test/jdk/tools/launcher/VersionCheck.java
@@ -23,7 +23,7 @@
 
 /**
  * @test
- * @bug 6545058 6611182 8016209 8139986 8162746
+ * @bug 6545058 6611182 8016209 8139986 8162746 8278967
  * @summary validate and test -version, -fullversion, and internal, as well as
  *          sanity checks if a tool can be launched.
  * @modules jdk.compiler
@@ -126,9 +126,12 @@ public class VersionCheck extends TestHelper {
     static String getVersion0(boolean allLines, String... argv) {
         TestHelper.TestResult tr = doExec(argv);
         StringBuilder out = new StringBuilder();
-        // remove the HotSpot line
+        // remove the HotSpot line and security manager deprecation warnings
         for (String x : tr.testOutput) {
-            if (allLines || !x.matches(".*Client.*VM.*|.*Server.*VM.*")) {
+            if (allLines || !x.matches(".*Client.*VM.*|" +
+                                       ".*Server.*VM.*|" +
+                                       "WARNING:.*terminally.*deprecated.*|" +
+                                       "WARNING:.*System::setSecurityManager.*")) {
                 out = out.append(x + "\n");
             }
         }


dholmes-ora (changes requested):
Hi Stuart,

I think specifying "allow" would be the behaviour preserving change here. That avoids any risk that enabling the SM earlier changes some behaviour during VM initialization.

Cheers,
David

stuart-marks:
> I think specifying "allow" would be the behaviour preserving change here.

This is strictly true. I did think about doing this. However, I don't think it makes much of a practical difference. It's always been fully supported to enable the security manager using properties, and the rmiregistry main hardly does anything before enabling the security manager. As long as we're going to the trouble of setting a property to _allow_ the security manager, might as well _enable_ it at the same time, I think.

In addition, the warning messages emitted by enabling the security manager with the API are worse than those from the command line. Using the API gives this:

```
WARNING: A terminally deprecated method in java.lang.System has been called
WARNING: System::setSecurityManager has been called by sun.rmi.registry.RegistryImpl
WARNING: Please consider reporting this to the maintainers of sun.rmi.registry.RegistryImpl
WARNING: System::setSecurityManager will be removed in a future release
```

and using the property gives this:

```
WARNING: A command line option has enabled the Security Manager
WARNING: The Security Manager is deprecated and will be removed in a future release
```

Using the property gives a shorter warning and one that is less misleading (!) than the one from use of the API. (I mean, it's kind of ridiculous for the `rmiregistry` command to emit a message suggesting that something be reported to the maintainers of `RegistryImpl`. In some sense it would be good if there were a way to disable the warning, but I rate the likelihood of that occurring as low.)

dholmes-ora:
My concern is that having the SM installed during part of VM initialization could lead to different behaviour compared to installing the SM after that. This may be a small risk but it is a risk, and not something we are likely to observe in our own testing.

If there is no way to hide those warning messages (@SuppressWarnings in the code?) then we have a serious flaw in our warning system. As you say these warnings are for the developer not the end user!

AlanBateman:
As things stand, `rmiregsitry -J-Djava.security.manager` and `rmiregistry -J-Djava.security.manager=allow` are equivalent because rmiregistry sets the default SM. Some difference may be observed if someone is running rmiregistry with a custom system class loader set, or custom file system provider, or running it with a JVM TI agent that is looking at events between VMStart and VMInit but these seem somewhat obscure scenarios for a command line program If rmiregstry worked with ToolProvider then I think there would be more to discuss. If the final patch is to have the launcher set the default security manager then we may want to change RegistryImpl.createRegistry to fail if not set.

The warning that is emitted for both cases is expected. JEP 411 is very clear that it there is no mechanism to suppress it. We may need to add a release note to document that rmiregistry will emit a warning a startup, that will at least be something to point at in the event that anyone asks or reports an issue.

stuart-marks:
I think there's little to worry about with custom configurations of the `rmiregistry` tool. What somebody might do is to enable a customized security manager using properties... of course they'd also have to ensure that the customized SM is somewhere in the classpath as well, by adding more command line options. That seems fairly unlikely to me. Anyone who wants to run an RMI registry service in some specialized configuration would probably just set things up themselves and then use the `LocateRegistry.createRegistry` API instead of trying to tinker with the `rmiregistry` command.

If `rmiregistry` enables the SM using properties, then yes we can probably change the code to assert that the SM is running instead of conditionally enabling it like it does now.

Next headache is that `tools/launcher/VersionCheck.jtr` fails because it sees the warning messages instead of the version string. (Argh!) Interestingly this test passes even though `rmiregistry` currently fails to operate normally, because it runs well enough to print its version string, but not well enough to start up a registry service. (Double argh!)

The warnings policy is a separate issue being discussed elsewhere.

dholmes-ora:
In the same kind of PR (https://github.com/openjdk/jdk18/pull/53) for `jstatd` @AlanBateman writes:
> This looks okay in that it is now worked exactly as it did in JDK 17.

And that PR uses `allow` as I have suggested here (to preserve existing behaviour). All the affected launchers should be fixed in the same consistent way.

stuart-marks:
Enabling the security manager using a property, versus setting the property to `allow` and enabling it in code, is mostly a distinction without a difference. I don't think there is really a need to have the different tools be consistent. Local tool considerations probably swamp the need for any cross-tool consistency.

In this case some of the RMI registry tests set the property to `allow` on the command line and then rely on the code to enable the security manager using the API, so it's much easier to switch the `rmiregistry` launcher to use the `allow` technique instead. This is sort-of the tail (the tests) wagging the doc (the tool) but there doesn't seem to any benefit to be gained from fiddling around with the tests and the RMI test library.

I've also updated VersionCheck.java to filter out the security manager warning messages.

AlanBateman (approved):
This version looks okay, avoids having an attempt to set the SM in createRegistry always be skipped.

erikj79 (approved):
Build change looks good.

PR #46, opened 2021-12-18
8278948: compiler/vectorapi/reshape/TestVectorCastAVX1.java crashes in assembler

merykitty:
This patch fixes a crash spotted in `compiler/vectorapi/reshape/TestVectorCastAVX1.java` in mainline. The reason for the failure is the incorrect vector encoding of integer promotion operation leads to unsupported instruction `vpmovsxbd/vpmovsxwd ymm, xmm` on AVX1. For the same reason we currently cannot cast a short or byte vector to a 256-bit float vector on AVX1, so I also fixed that.

---------
### Progress

### Issue
 * [JDK-8278948](https://bugs.openjdk.java.net/browse/JDK-8278948): compiler/vectorapi/reshape/TestVectorCastAVX1.java crashes in assembler

### Reviewers
 * [Vladimir Kozlov](https://openjdk.java.net/census#kvn) (@vnkozlov - **Reviewer**) ⚠️ Review applies to 5afabe5b5ef0fe737a3bcd81b12b54f25e092892

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/46/head:pull/46` \
`$ git checkout pull/46`

`$ git checkout pull/46` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/46/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 46`

`$ git pr show -t 46`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/46.diff">https://git.openjdk.java.net/jdk18/pull/46.diff</a>

</details>

### DIFF
--- a/src/hotspot/cpu/x86/x86.ad
+++ b/src/hotspot/cpu/x86/x86.ad
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2011, 2021, Oracle and/or its affiliates. All rights reserved.
+// Copyright (c) 2011, 2022, Oracle and/or its affiliates. All rights reserved.
 // DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
 //
 // This code is free software; you can redistribute it and/or modify it
--- a/src/hotspot/cpu/x86/x86.ad
+++ b/src/hotspot/cpu/x86/x86.ad
@@ -1773,17 +1773,9 @@ const bool Matcher::match_rule_supported_vector(int opcode, int vlen, BasicType 
       }
       break;
     case Op_VectorCastB2X:
-      if (size_in_bits == 256 && UseAVX < 2) {
-        return false; // Implementation limitation
-      }
-      break;
     case Op_VectorCastS2X:
-      if (is_integral_type(bt) && size_in_bits == 256 && UseAVX < 2) {
-        return false;
-      }
-      break;
     case Op_VectorCastI2X:
-      if (is_integral_type(bt) && size_in_bits == 256 && UseAVX < 2) {
+      if (bt != T_DOUBLE && size_in_bits == 256 && UseAVX < 2) {
         return false;
       }
       break;
@@ -6896,11 +6888,12 @@ instruct vcastBtoX(vec dst, vec src) %{
       case T_LONG:
         __ vpmovsxbq($dst$$XMMRegister, $src$$XMMRegister, vlen_enc);
         break;
-      case T_DOUBLE:
-        __ vpmovsxbd($dst$$XMMRegister, $src$$XMMRegister, vlen_enc);
+      case T_DOUBLE: {
+        int mid_vlen_enc = (vlen_enc == Assembler::AVX_512bit) ? Assembler::AVX_256bit : Assembler::AVX_128bit;
+        __ vpmovsxbd($dst$$XMMRegister, $src$$XMMRegister, mid_vlen_enc);
         __ vcvtdq2pd($dst$$XMMRegister, $dst$$XMMRegister, vlen_enc);
         break;
-
+      }
       default: assert(false, "%s", type2name(to_elem_bt));
     }
   %}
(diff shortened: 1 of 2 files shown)


vnkozlov (commented):
@merykitty  please, add comment in bug report how to reproduced it and how failure looks like.
We do run AVX1 testing with these tests and I did not see failure.

@jatin-bhateja or @sviswa7 please look on it.

merykitty:
Hi, I have commented in the bug report, the crash can be observed by uncommenting line 51 in `compiler/vectorapi/reshape/utils/TestCastMethods.java` which tells `TestVectorCastAVX1.java` to perform a cast from `Short64Vector` to `Double256Vector`.

https://github.com/openjdk/jdk/blob/f7309060ded0edb1e614663572f876d83b77c28e/test/hotspot/jtreg/compiler/vectorapi/reshape/utils/TestCastMethods.java#L51

The crash happens at `void Assembler::vpmovsxwd(XMMRegister dst, XMMRegister src, int vector_len)` due to `assert(vector_len == AVX_128bit ? VM_Version::supports_avx() : vector_len == AVX_256bit ? VM_Version::supports_avx2() :  VM_Version::supports_evex(), "");`

Thank you very much.

vnkozlov:
Thank you for explaining. These tests are not present in JDK18 and that confused me. I suggest to add new simple regression test with your changes for JDK 18 so we can verify the fix. And file a separate RFE for JDK 19 to uncomment lines in TestCastMethods.java when the fix is auto-pushed into JDK 19. 

And what about commented line 42 `makePair(BSPEC64, DSPEC256)1` ?

src/hotspot/cpu/x86/x86.ad
```diff
@@ -1773,17 +1773,9 @@ const bool Matcher::match_rule_supported_vector(int opcode, int vlen, BasicType
...
-      }
-      break;
     case Op_VectorCastS2X:
-      if (is_integral_type(bt) && size_in_bits == 256 && UseAVX < 2) {
-        return false;
-      }
-      break;
     case Op_VectorCastI2X:
-      if (is_integral_type(bt) && size_in_bits == 256 && UseAVX < 2) {
+      if (bt != T_DOUBLE && size_in_bits == 256 && UseAVX < 2) {
```
sviswa7, line 1778: CastI2X  should work with (UseAVX == 1) for (bt == T_FLOAT) so the prior code was correct for CastI2X. 
The fix is only needed for  CastS2X and CastB2X.

merykitty:
Hi @vnkozlov , I have added tests for this regression.
The cast from byte to double also suffers this miscalculation, it is also mistakenly rejected from intrinsification before and I also fix it here.
Thank you very much.

src/hotspot/cpu/x86/x86.ad
merykitty: AFAIK int vectors of size 256 bit are not supported on AVX1 so we cannot cast from int vectors to float vectors of size 256 anyway. So it is essentially the same.

vnkozlov: x86 docs say that `vcvtdq2ps` is AVX1 instruction:
image

vnkozlov:
> Hi @vnkozlov , I have added tests for this regression.
> The cast from byte to double also suffers this miscalculation, it is also mistakenly rejected from intrinsification before and I also fix it here.
> Thank you very much.

Thank you for adding test.
Please, address Sandhya's comment about 256 bit CastI2F support with AVX1. I added snapshot from x86 doc to prove it.

src/hotspot/cpu/x86/x86.ad
merykitty: Hi, while the instruction is supported in hardware, the VM does not support the vector shape. As a result, there should not be a situation where we have a vector node with that shape on AVX1. Thank you very much.

vnkozlov: Why do you think that? What about this code when length is the same (and UseAVX could be 1):
https://github.com/openjdk/jdk18/blob/master/src/hotspot/cpu/x86/x86.ad#L7069

merykitty: Because the input of `VectorCastI2XNode` would be a vector node of type int and size 256, this shape is not supported on AVX1. So, we should not have this node because its input should not appear.

vnkozlov: Right.

vnkozlov (commented):
All questions were answered.
I will run testing before approval.

vnkozlov (approved):
Testing tier1-3 passed.

vnkozlov:
@sviswa7 please, review final changes so that it could be pushed.

To "close loop" on vector size limits. Here is where it set:
https://github.com/openjdk/jdk18/blob/master/src/hotspot/cpu/x86/x86.ad#L2068
As @merykitty correctly pointed AVX1 supports 256 bits Float vectors but not Integer vectors.

vnkozlov:
@merykitty please, merge and update copyright year to 2022 in changed files.

sviswa7:
Yes, we don't support 256 bit integral vectors in the JVM when UseAVX=1. The patch looks good to me.

sviswa7 (approved):

merykitty:
Thank you very much for the reviews, I have updated the copyright year to 2022 already.
/integrate

PR #49, opened 2021-12-20
8278889: AArch64: [vectorapi] VectorMaskLoadStoreTest.testMaskCast() test fail

e1iu:
This bug appears intermittently and it's caused by vmaskAll_immI[1]
when the vector mask size is smaller than max predicate size of running
machine. It generates an all-true predicate without considering those
inactive bits. That may result in the wrong result of VectorMask.toLong.
The problematic code is as below:

```
        ShortVector.SPECIES_64.MaskAll(true).toLong()

assembly:

        ptrue   p0.h          <= MaskAll(true)
        mov     z16.h, p0/z, #1
        mov     z17.h, #0
        uzp1    z16.b, z16.b, z17.b
        fmov    x10, d16
        orr     x10, x10, x10, lsr #7
        orr     x10, x10, x10, lsr #14
        orr     x10, x10, x10, lsr #28
        and     x10, x10, #0xff

(gdb) p/x $p0 # on an SVE machine with vector length as 64 in bytes
$1 = {0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55}

Expected:
(gdb) p/x $p0
$1 = {0x55, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00}

```

Considering MaskAll is used in VectorMask.fromLong() only for a special
case and relies on the mechanism of inline and intrinsification, even it
could be optimized out, this patch also adds test cases for MaskAll to
reproduce this issue stably.

Also fix a small issue on register utilization for
sve_reduce_[max|min][D|F].

[1] https://github.com/openjdk/jdk18/blob/master/src/hotspot/cpu/aarch64/aarch64_sve.ad#L416

hotspot/compiler/vectorapi, jdk/incubator/vector passed on SVE enabled
system.

Change-Id: I9631f26f9232ffe7a28b74f14062d945c32fa1fb

---------
### Progress

### Issue
 * [JDK-8278889](https://bugs.openjdk.java.net/browse/JDK-8278889): AArch64: [vectorapi] VectorMaskLoadStoreTest.testMaskCast() test fail

### Reviewers
 * [Ningsheng Jian](https://openjdk.java.net/census#njian) (@nsjian - Committer) ⚠️ Review applies to 109d88384faf0af0f821c7744da92ed420313c9d

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/49/head:pull/49` \
`$ git checkout pull/49`

`$ git checkout pull/49` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/49/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 49`

`$ git pr show -t 49`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/49.diff">https://git.openjdk.java.net/jdk18/pull/49.diff</a>

</details>


src/hotspot/cpu/aarch64/aarch64_sve_ad.m4
```diff
@@ -377,19 +378,21 @@ instruct vmaskAll$1(pRegGov dst, ifelse($1, `I', iRegIorL2I, iRegL) src, vReg tm
...
   format %{ "sve_dup $tmp, $src\n\t"
-            "sve_cmpne $dst, $tmp, 0\t# mask all (sve) ($2)" %}
+            "sve_ptrue_lanecnt $dst\n\t"
+            "sve_cmpne $dst, $dst, $tmp, 0\t# mask all (sve) ($2)" %}
   ins_encode %{
     BasicType bt = Matcher::vector_element_basic_type(this);
     Assembler::SIMD_RegVariant size = __ elemType_to_regVariant(bt);
     __ sve_dup(as_FloatRegister($tmp$$reg), size, as_Register($src$$reg));
-    __ sve_cmp(Assembler::NE, as_PRegister($dst$$reg), size, ptrue, as_FloatRegister($tmp$$reg), 0);
+    __ sve_ptrue_lanecnt(as_PRegister($dst$$reg), size, Matcher::vector_length(this));
```
nsjian: I think you can generate this insn conditionally, only when current vector size is not MaxVectorSize.

nsjian (approved):
Thanks for the fix! Overall looks good to me. Just one enhancement suggestion.

src/hotspot/cpu/aarch64/aarch64_sve_ad.m4
e1iu: Fixed. Thanks for your review.

src/hotspot/cpu/aarch64/c2_MacroAssembler_aarch64.cpp
```diff
@@ -1232,3 +1232,39 @@ void C2_MacroAssembler::sve_reduce_integral(int opc, Register dst, BasicType bt,
...
+      sve_ptrue(dst, size, /* VL32 */ 0b01010);
+      break;
+    case 64:
+      sve_ptrue(dst, size, /* VL64 */ 0b01011);
+      break;
+    case 128:
+      sve_ptrue(dst, size, /* VL128 */ 0b01100);
+      break;
+    case 256:
+      sve_ptrue(dst, size, /* VL256 */ 0b01101);
```
dean-long, line 1264: Why not use an enum for these magic constants?

vnkozlov (changes requested):
I ran tier1 and 3 tests in `test/hotspot/jtreg/gc/stress/gcbasher/` failed on macosx-aarch64:

```
java.lang.IllegalStateException
	at gc.stress.gcbasher.ByteCursor.readUtf8(ByteCursor.java:110)
	at gc.stress.gcbasher.Decompiler.decodeConstantPool(Decompiler.java:310)
	at gc.stress.gcbasher.Decompiler.<init>(Decompiler.java:42)
	at gc.stress.gcbasher.TestGCBasher.parseClassFiles(TestGCBasher.java:46)
	at gc.stress.gcbasher.TestGCBasher.main(TestGCBasher.java:63)
	at gc.stress.gcbasher.TestGCBasherWithG1.main(TestGCBasherWithG1.java:40)
```

It seems new failure.

src/hotspot/cpu/aarch64/c2_MacroAssembler_aarch64.cpp
e1iu: I think these magic numbers are only used by sve_ptrue and probably there should be no one need to write these magic numbers again. If these magic numbers are really common used,  I agree with yours that enum is better.

e1iu:
> I ran tier1 and 3 tests in `test/hotspot/jtreg/gc/stress/gcbasher/` failed on macosx-aarch64:
> 
> ```
> java.lang.IllegalStateException
> 	at gc.stress.gcbasher.ByteCursor.readUtf8(ByteCursor.java:110)
> 	at gc.stress.gcbasher.Decompiler.decodeConstantPool(Decompiler.java:310)
> 	at gc.stress.gcbasher.Decompiler.<init>(Decompiler.java:42)
> 	at gc.stress.gcbasher.TestGCBasher.parseClassFiles(TestGCBasher.java:46)
> 	at gc.stress.gcbasher.TestGCBasher.main(TestGCBasher.java:63)
> 	at gc.stress.gcbasher.TestGCBasherWithG1.main(TestGCBasherWithG1.java:40)
> ```
> 
> It seems new failure.

Thanks for your testing! Looks like it's the same as https://bugs.openjdk.java.net/browse/JDK-8275263.

vnkozlov:
It is actually [8275316](https://bugs.openjdk.java.net/browse/JDK-8275316) which was closed as duplicate [8275263 ](https://bugs.openjdk.java.net/browse/JDK-8275263).
But, as I understand, 8275263 was also closed as duplicate of [BACKOUT] [8275262](https://bugs.openjdk.java.net/browse/JDK-8275262).
So there should not be code in JDK 18 which cause it.

Note, I ran additional testing of the same local repo build but without this **8278889** changes and I don't see `gcbasher` failures.

e1iu:
> It is actually [8275316](https://bugs.openjdk.java.net/browse/JDK-8275316) which was closed as duplicate [8275263 ](https://bugs.openjdk.java.net/browse/JDK-8275263). But, as I understand, 8275263 was also closed as duplicate of [BACKOUT] [8275262](https://bugs.openjdk.java.net/browse/JDK-8275262). So there should not be code in JDK 18 which cause it.
> 
> Note, I ran additional testing of the same local repo build but without this **8278889** changes and I don't see `gcbasher` failures.

Thanks for your feedback.

The failure is really bizarre. Is it possible that caused by the specific toolchain ? This patch doesn't change any common code except the test cases, all the changes are SVE related, and I think mac couldn't touch SVE related code at this moment. Moreover, the log is the same as [8275316](https://bugs.openjdk.java.net/browse/JDK-8275316), which means perhaps some other patches will trigger this failure as well before the root cause has been fixed.  In this way, expose this failure and post a new JBS to involve some mac experts maybe not bad. Otherwise, some patches which looks okay but would be blocked unfortunately.

vnkozlov:
This is indeed bizarre. I submitted new build in our infra only for macosx-aarch64-debug and test passed with it (20 times). And it immediately failed with old build.
I checked: OS version is the same and clang (and xcode) exactly the same for both builds.

My only suspicion is we may have some uninitialized value[s] in VM somewhere and depending on JVM's code and data layout we can get these strange failures.

I am repeating whole tier1 build and testing on all platforms.

vnkozlov:
I filed https://bugs.openjdk.java.net/browse/JDK-8279177
tier1 re-testing passed. Running tier2 and tier3 now.

vnkozlov (approved):
Re-testing passed.

e1iu:
Is there anyone can help to sponsor this if no more reviews?

e1iu:
Thanks for all reviewers:P

PR #50, opened 2021-12-20
8279011: JFR: JfrChunkWriter incorrectly handles int64_t chunk size as size_t

shipilev:
See the investigation in the bug.

Spot the problem:

```
int64_t JfrChunkWriter::write_chunk_header_checkpoint(bool flushpoint) {
   ...
   const size_t sz_written = size_written(); // <-- returns int64_t
   write_be_at_offset(sz_written, chunk_size_offset); // <--- template instantiation with type=size_t
   return sz_written;
 }
```

This would have been nearly fine -- small `size_t` -> `int64_t` conversion is okay value-wise. But `write_be_at_offset` calculates the position for the writeout using `sizeof(T)`, which silently borks the whole thing on at least 32-bit platforms, where `sizeof(size_t)` != `sizeof(int64_t)`. 

Additional testing:
 - [x] Linux x86_64 `jdk_jfr` (no regressions)
 - [x] Linux x86_32 `jdk_jfr` (many failing tests now pass)

---------
### Progress

### Issue
 * [JDK-8279011](https://bugs.openjdk.java.net/browse/JDK-8279011): JFR: JfrChunkWriter incorrectly handles int64_t chunk size as size_t

### Reviewers
 * [Markus Grönlund](https://openjdk.java.net/census#mgronlun) (@mgronlun - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/50/head:pull/50` \
`$ git checkout pull/50`

`$ git checkout pull/50` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/50/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 50`

`$ git pr show -t 50`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/50.diff">https://git.openjdk.java.net/jdk18/pull/50.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/jfr/recorder/repository/jfrChunkWriter.cpp
+++ b/src/hotspot/share/jfr/recorder/repository/jfrChunkWriter.cpp
@@ -207,7 +207,7 @@ int64_t JfrChunkWriter::write_chunk_header_checkpoint(bool flushpoint) {
   const u4 checkpoint_size = current_offset() - event_size_offset;
   write_padded_at_offset<u4>(checkpoint_size, event_size_offset);
   set_last_checkpoint_offset(event_size_offset);
-  const size_t sz_written = size_written();
+  const int64_t sz_written = size_written();
   write_be_at_offset(sz_written, chunk_size_offset);
   return sz_written;
 }


mgronlun:
Nice find.

But surprised there is not a compile error here (both incompatible type size and sign), if sizeof(size_t) == 4 and sizeof(int64_t) == 8:

const size_t sz_written = size_written(); // <-- returns int64_t

So the value is written as a 4-byte quantity only, since it is dimensioned using sizeof(size_t), and therefore does not fill the full 8- bytes slot associated with the 'chunksize' header field. And uninitialized memory in the high 4-bytes leads to this high value?

shipilev:
> But surprised there is not a compile error here (both incompatible type size and sign), if sizeof(size_t) == 4 and sizeof(int64_t) == 8:
> 
> const size_t sz_written = size_written(); // <-- returns int64_t

I believe narrowing conversions like these are implicit in C++, and do not produce warnings in modern compilers. See: https://godbolt.org/z/48oh91YE9
 
> So the value is written as a 4-byte quantity only, since it is dimensioned using sizeof(size_t), and therefore does not fill the full 8- bytes slot associated with the 'chunksize' header field. And uninitialized memory in the high 4-bytes leads to this high value?

Actually, I think it is worse than that, since `sizeof(T)*len + len` can produce the `pos` that is way off, because of multiplication by `len`:

```
template <typename BE, typename IE, typename WriterPolicyImpl >
template <typename T>
inline void WriterHost<BE, IE, WriterPolicyImpl>::be_write(const T* value, size_t len) {
  assert(value != NULL, "invariant");
  assert(len > 0, "invariant");
  // Might need T + 1 size
  u1* const pos = ensure_size(sizeof(T) * len + len);
  if (pos) {
    this->set_current_pos(BE::be_write(value, len, pos));
  }
```

So it might just store the `chunk_size` somewhere else...

mgronlun:
Len is only internally derived to ensure the size to be written can be accommodated by the internal buffer system, it does not affect the position of where to write on disk.

mgronlun (approved):
Looks good. Thanks.

shipilev:
> Len is only internally derived to ensure the size to be written can be accommodated by the internal buffer system, it does not affect the position of where to write on disk.

All right. Then it must be just uninitialized 32-bit head/tail in 64-bit value. In the bug report, we see `9896023423722239`, which is `0x23286180dc22ff`. Out of which, higher 32-bits `0x232861` looks like a real chunk size, `2304097`. This is similar to the chunk size 64-bit VM reports in the same test.

shipilev:
Any other (R)eviewers needd? I think this is pretty trivial.

mgronlun:
> Any other (R)eviewers needd? I think this is pretty trivial.

Please go ahead Aleksey, thank you.

shipilev:
> Please go ahead Aleksey, thank you.

Thanks. I followed up on GHA failures, those seem to be unrelated (Shenandoah and compiler bugs in JDK 18).

/integrate

PR #51, opened 2021-12-20
8272058: 25 Null pointer dereference defect groups in 4 files

dcubed-ojdk:
A small refactoring to resolve a Parfait complaint about the return value from
`MacroAssembler::target_addr_for_insn(address insn_addr, unsigned insn)`
on AARCH64. The logic that supports returning `nullptr` as the target addr for
a particular instruction is moved from
`MacroAssembler::target_addr_for_insn(address insn_addr, unsigned insn)` to
`MacroAssembler::target_addr_for_insn_allow_null_ret(address insn_addr, unsigned insn)`.
A couple of `target_addr_for_insn()` call sites that can tolerate a `nullptr` are
converted to use `target_addr_for_insn_allow_null_ret()`.

This fix has been tested with Mach5 Tier[1-3].

---------
### Progress

### Issue
 * [JDK-8272058](https://bugs.openjdk.java.net/browse/JDK-8272058): 25 Null pointer dereference defect groups in 4 files

### Reviewers
 * [Gerard Ziemski](https://openjdk.java.net/census#gziemski) (@gerard-ziemski - Committer) ⚠️ Review applies to 4c93290ecd205de857c07999dad904e098e9d8d6
 * [Coleen Phillimore](https://openjdk.java.net/census#coleenp) (@coleenp - **Reviewer**) ⚠️ Review applies to 0c23024fc117be69b6406602f40bee3e8f7be3d3
 * [Dean Long](https://openjdk.java.net/census#dlong) (@dean-long - **Reviewer**) ⚠️ Review applies to 0c23024fc117be69b6406602f40bee3e8f7be3d3

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/51/head:pull/51` \
`$ git checkout pull/51`

`$ git checkout pull/51` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/51/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 51`

`$ git pr show -t 51`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/51.diff">https://git.openjdk.java.net/jdk18/pull/51.diff</a>

</details>


dcubed-ojdk:
The Parfait complaint at the root of this issue is:

MacroAssembler::target_addr_for_insn
open/src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp

```
286. | } else if (Instruction_aarch64::extract(insn, 31, 22) == 0b1011100101 &&
287. |            Instruction_aarch64::extract(insn, 4, 0) == 0b11111) {
288. |   return 0;  // Null pointer introduced .
289. | } else {
290. |   ShouldNotReachHere();
```

dholmes-ora:
Hi Dan,

In terms of the refactoring and addressing the warning this seems reasonable. But in terms of semantics this claims that all the remaining callsites for `target_addr_for_insn` can never pass an `insn` that can lead to a NULL address. I can't verify that claim. This needs some Aarch64 experts to chime in - perhaps @theRealAph?

Thanks,
David

dcubed-ojdk:
@dholmes-ora - Thanks for taking a look. I did take the time to look at all the
remaining call sites for `target_addr_for_insn` and, in some cases, their callers
and I didn't see anything where it made sense to run into an `insn` that could
lead to a `nullptr` address. Of course, that's just code inspection on my part
so I also created `guarantee()` traps at all the call-sites and ran those bits
thru Mach5 Tier[1-8] testing without hitting any of the traps. As you pointed
out in the bug report, that kind of testing isn't proof either, but it helps me
believe that I didn't miss anything too obvious.

Having @theRealAph chime in here would be wonderful!

dcubed-ojdk:
@gerard-ziemski - When you have a chance, could you take a look at this one?

gerard-ziemski (approved):
I don't understand the underlying code, but from a big picture point of view, splitting the call into "null not allowed" and "null allowed" makes sense. If the testing came back fine, then that's the best we can do, without asking someone who wrote this code to chime in.

Just one question: why do we need the "null allowed" version here?

void NativeMovRegMem::verify() {
 #ifdef ASSERT
   address dest = MacroAssembler::target_addr_for_insn_allow_null_ret(instruction_address());
 #endif
 }

dcubed-ojdk:
@gerard-ziemski - Thanks for the review.

According to "git blame", the original author is:

9c458decf5f1 hotspot/src/cpu/aarch64/vm/macroAssembler_aarch64.cpp (Andrew Haley           2015-01-20 11:34:17 -0800  211) address MacroAssembler::target_addr_for_insn(address insn_addr, unsigned insn) {

which I believe is why @dholmes-ora pinged @theRealAPH on this PR...

As for this function:

```
void NativeMovRegMem::verify() {
#ifdef ASSERT
  address dest = MacroAssembler::target_addr_for_insn_allow_null_ret(instruction_address());
#endif
}
```

I did not see any harm in converting to `target_addr_for_insn_allow_null_ret`.
The baseline code worked just fine before I refactored it so I'm giving
the function equivalence. In other words, if `NativeMovRegMem::verify()`
was previous called with an `instruction_address()` value that leads to
a nullptr, then baseline `verify()` would work without error. I wanted the
new `verify()` to also work the same.

gerard-ziemski:
> @gerard-ziemski - Thanks for the review.
> 
> According to "git blame", the original author is:
> 
> [9c458de](https://github.com/openjdk/jdk18/commit/9c458decf5f1ed468639de57f9f34adab98a593d) hotspot/src/cpu/aarch64/vm/macroAssembler_aarch64.cpp (Andrew Haley 2015-01-20 11:34:17 -0800 211) address MacroAssembler::target_addr_for_insn(address insn_addr, unsigned insn) {
> 
> which I believe is why @dholmes-ora pinged @theRealAph on this PR...
> 
> As for this function:
> 
> ```
> void NativeMovRegMem::verify() {
> #ifdef ASSERT
>   address dest = MacroAssembler::target_addr_for_insn_allow_null_ret(instruction_address());
> #endif
> }
> ```
> 
> I did not see any harm in converting to `target_addr_for_insn_allow_null_ret`. The baseline code worked just fine before I refactored it so I'm giving the function equivalence. In other words, if `NativeMovRegMem::verify()` was previous called with an `instruction_address()` value that leads to a nullptr, then baseline `verify()` would work without error. I wanted the new `verify()` to also work the same.

As long as parfait doesn't complain here - I didn't see where we checked for NULL explicitly in this case.  I assume that parfait skips the ASSERT parts?

dcubed-ojdk:
> As long as parfait doesn't complain here - I didn't see where we checked for NULL
> explicitly in this case. I assume that parfait skips the ASSERT parts?

Parfait does not complain anymore. The key about this call is that we don't de-reference
the return value from `target_addr_for_insn_allow_null_ret`. So we allow nullptr to be
returned, but we don't use it.

dholmes-ora:
Perhaps Boris Ulasevich <boris.ulasevich@bell-sw.com> could review?

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -283,15 +283,20 @@ address MacroAssembler::target_addr_for_insn(address insn_addr, unsigned insn) {
...
-             Instruction_aarch64::extract(insn, 4, 0) == 0b11111) {
-    return 0;
   } else {
     ShouldNotReachHere();
   }
   return address(((uint64_t)insn_addr + (offset << 2)));
 }
 
+address MacroAssembler::target_addr_for_insn_allow_null_ret(address insn_addr, unsigned insn) {
+  if (Instruction_aarch64::extract(insn, 31, 22) == 0b1011100101 &&
```
bulasevich: Consider replacing raw constants check with NativeInstruction::is_ldrw_to_zr function call which performs the same check.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
```diff
@@ -283,15 +283,20 @@ address MacroAssembler::target_addr_for_insn(address insn_addr, unsigned insn) {
     return address(uint64_t(Instruction_aarch64::extract(insns[0], 20, 5))
                    + (uint64_t(Instruction_aarch64::extract(insns[1], 20, 5)) << 16)
                    + (uint64_t(Instruction_aarch64::extract(insns[2], 20, 5)) << 32));
-  } else if (Instruction_aarch64::extract(insn, 31, 22) == 0b1011100101 &&
```
bulasevich, line 287: It must be a check for safePoint poll: ldrw zr, polling_page. Technically, the target address for this instruction is zero and the function is correct. I am not sure Parfait is right in his complaints. After all, does this code ever works? That is, if you remove it, will the SholdNotReachHere below fire?

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
dcubed-ojdk: Good catch. There are actually two uses of the raw constants in
macroAssembler_aarch64.cpp and I've cleaned up both.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
dcubed-ojdk: I don't think the safepoint poll is coming thru this code path. I ran this fix
through Mach5 Tier[1-6] and did not see any failures caused by hitting
a `ShouldNotReachHere()` call. It's hard to imagine that Tier[1-6] testing
wouldn't exercise the safepoint polling code paths.

I've also verified that the Parfait tool no longer issues the complaints.

I will poke around and see if I can determine what the safepoint polling
is doing.

dcubed-ojdk: @bulasevich - I've made another pass through the code and checked the
calls to `MacroAssembler::target_addr_for_insn()` and I don't see any of
those calls in areas related to safepoint polling. From the other POV, I
took a look at the aarch64 safepoint polling code and I don't see a way
into `MacroAssembler::target_addr_for_insn()` .

It is entirely possible that I'm missing something here and would appreciate
it if you could sanity check my conclusions here.

dcubed-ojdk:
Mach5 Tier[1-3] testing on v02 is clean (no failures).

Update: Mach5 Tier[4-6] test on v02 is clean (1 known, unrelated client test failure).

dcubed-ojdk:
@bulasevich and/or @theRealAph - pinging for reviews...

dcubed-ojdk:
@bulasevich and/or @theRealAph - pinging for reviews...

dcubed-ojdk:
@bulasevich - Please re-review this PR. The JDK18 RDP2 cut off is next week (01.20) and
I would like to have this PR integrated before then.

@theRealAph - Since I'm changing code that you integrated, I would prefer to hear from
you on this PR. Please review these changes soon. The JDK18 RDP2 cut off is next week
(01.20) and I would like to have this PR integrated before then.

dcubed-ojdk:
Mach5 Tier[1-3] on the latest bits had no failures.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
ghost: I believe this code is dead, in both `MacroAssembler::pd_patch_instruction()` and `MacroAssembler::target_addr_for_insn()`, since the poll read is PIC, we are not patching and should not need to acquire any target to update.

The associated `fix_relocation_after_move()` will not invoke any of the above on the poll read since the `maybe_cpool_ref()` condition is used:
```
void poll_Relocation::fix_relocation_after_move(const CodeBuffer* src, CodeBuffer* dest) {
  if (NativeInstruction::maybe_cpool_ref(addr())) {
    address old_addr = old_addr_for(addr(), src, dest);
    MacroAssembler::pd_patch_instruction(addr(), MacroAssembler::target_addr_for_insn(old_addr));
  }
}
```
Should be sufficient to simply remove these cases, possibly you could move an assert into the "unreachable" branch (for "precision").
```
} else {
  assert(!NativeInstruction::is_ldrw_to_zr(<addr>), "Unexpected poll read");
  ShouldNotReachHere();
}
```
(The code above in `fix_relocation_after_move()` might also be dead...)

Chip in @theRealAph ?

coleenp (approved):
I walked through the changes that you made in my local code base and by inspection and your testing, this change looks good.  Thanks for fixing the problem.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
coleenp: This was a nice refactoring.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
dcubed-ojdk: @phedlin - Thanks for chiming in on this review thread.

dcubed-ojdk:
@coleenp - Thanks for the review and the approval.

dean-long:
I'm glad you didn't have to change all the Got and Plt call sites that parfait was complaining about.  It seems to be all dead code left over from AOT.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.hpp
```diff
@@ -606,10 +606,15 @@ class MacroAssembler: public Assembler {
   static bool uses_implicit_null_check(void* address);
 
   static address target_addr_for_insn(address insn_addr, unsigned insn);
+  static address target_addr_for_insn_allow_null_ret(address insn_addr, unsigned insn);
   static address target_addr_for_insn(address insn_addr) {
```
dean-long, line 610: The or_null suffix that we use in shared code reads a little better to me.

dean-long (approved):

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
bulasevich: Good. Thank you!

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
bulasevich: I wondered if target_addr_for_insn ever gets the ldrw_to_zr instruction. 
Anyway, I am Ok to move it to target_addr_for_insn_allow_null_ret (target_addr_for_insn_or_null?).
Thank you.

bulasevich:
The change looks good for me (I'm not a reviewer). Thank you.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
theRealAph: Since we moved to using per-thread safepoints, I don't think we'll ever see a reloc against a safepoint poll. I've just looked through the AArch64 back end, and as far as I can see the old safepoint code is gone.

theRealAph: > @bulasevich - I've made another pass through the code and checked the calls to `MacroAssembler::target_addr_for_insn()` and I don't see any of those calls in areas related to safepoint polling. From the other POV, I took a look at the aarch64 safepoint polling code and I don't see a way into `MacroAssembler::target_addr_for_insn()` .
> 
> It is entirely possible that I'm missing something here and would appreciate it if you could sanity check my conclusions here.

I think you're right. As far as I can see, this code is dead since 
8253180: ZGC: Implementation of JEP 376: ZGC: Concurrent Thread-Stack Processing

which took out the code that generated safepoint relocs, but not the code here that relocated them.

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
theRealAph: Good.

dcubed-ojdk:
@dean-long - Thanks for the review and for the rename suggestion. I'm making
that change. Retesting the build now.

@bulasevich - Thanks for the re-review!

@theRealAph - Thanks for chiming in on the review thread.

dcubed-ojdk:
This latest version has been tested with a Mach5 Tier1 job set.
No failures in either linux-aarch64 or macosx-aarch64.

/integrate

src/hotspot/cpu/aarch64/macroAssembler_aarch64.cpp
dcubed-ojdk: I've filed the following new RFE to track the dead compiler code:

    JDK-8280283 dead compiler code found during the JDK-8272058 code review
    https://bugs.openjdk.java.net/browse/JDK-8280283

PR #68, opened 2021-12-23
8268297: jdk/jfr/api/consumer/streaming/TestLatestEvent.java times out

egahlin:
Hi,

Could I have a review of a fix that prevents a recording stream from dropping events/chunks. 

This occurs when rotation happens quickly, as in TestLatestEvent.java, and the clock is not making progress, on Windows sometimes. The result is a chunk header with duration 0 ns, meaning there could be two chunks with the same start timestamp, which confuses the RecordingStream when determining iteration order.

The fix is to always increase the elapsed time by at least 1 ns. There were similar code prior to JDK 17, but it was removed because a mechanism was added in Java that makes sure a chunk always gets a unique Instant.now() timestamp. Unfortunately, that timestamp does not always correlate with what the JVM writes, which leads to intermittent failures in 2-4% of the time. 

I also updated the test, so it can provide better information in case of new failures.

Testing:  
test/jdk/jdk/jfr/api/consumer/streaming/TestLatestEvent.java 600 times without failure on Windows
test/jdk/jdk/jfr/ 6 times without failure on Mac, Linux and Windows.

Thanks
Erik

---------
### Progress

### Issue
 * [JDK-8268297](https://bugs.openjdk.java.net/browse/JDK-8268297): jdk/jfr/api/consumer/streaming/TestLatestEvent.java times out

### Reviewers
 * [Markus Grönlund](https://openjdk.java.net/census#mgronlun) (@mgronlun - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/68/head:pull/68` \
`$ git checkout pull/68`

`$ git checkout pull/68` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/68/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 68`

`$ git pr show -t 68`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/68.diff">https://git.openjdk.java.net/jdk18/pull/68.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/jfr/recorder/repository/jfrChunk.cpp
+++ b/src/hotspot/share/jfr/recorder/repository/jfrChunk.cpp
@@ -47,6 +47,8 @@ static jlong nanos_now() {
   const jlong now = seconds * 1000000000 + nanos;
   if (now > last) {
     last = now;
+  } else {
+    ++last;
   }
   return last;
 }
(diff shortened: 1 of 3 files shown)


mgronlun (approved):
Looks good.

PR #70, opened 2021-12-23
8279222: Incorrect legacyMap.get in java.security.Provider after JDK-8276660

shipilev:
SonarCloud reports:
 A "Map<ServiceKey, Service>" cannot contain a "String" in a "ServiceKey" type.

```
     // clean up old alias if present
     Service prevAliasService = legacyMap.get(aliasAlg);
```

Should be `aliasKey`, like other accesses to `legacyMap`. This code is introduced by [JDK-8276660](https://bugs.openjdk.java.net/browse/JDK-8276660), so it affects JDK 18.

Additional testing:
 - [x] Linux x86_64 fastdebug `jdk_security`

---------
### Progress

### Issue
 * [JDK-8279222](https://bugs.openjdk.java.net/browse/JDK-8279222): Incorrect legacyMap.get in java.security.Provider after JDK-8276660

### Reviewers
 * [Alan Bateman](https://openjdk.java.net/census#alanb) (@AlanBateman - **Reviewer**)
 * [Valerie Peng](https://openjdk.java.net/census#valeriep) (@valeriepeng - **Reviewer**)

### Contributors
 * Valerie Peng `<valeriep@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/70/head:pull/70` \
`$ git checkout pull/70`

`$ git checkout pull/70` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/70/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 70`

`$ git pr show -t 70`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/70.diff">https://git.openjdk.java.net/jdk18/pull/70.diff</a>

</details>


src/java.base/share/classes/java/security/Provider.java
```diff
@@ -1149,7 +1149,7 @@ private void parseLegacy(String name, String value, OPType opType) {
             switch (opType) {
                 case ADD:
                     // clean up old alias if present
-                    Service prevAliasService = legacyMap.get(aliasAlg);
+                    Service prevAliasService = legacyMap.get(aliasKey);
```
AlanBateman, line 1152: The change looks okay but I'm surprised this wasn't caught by tests. 
@valeriepeng - would it be feasible to have a test that removes an old alias so that this code is exercised by tests?

valeriepeng: Hmm, this "ADD" case should be covered by existing regression tests. I will take a look. Thanks.

valeriepeng: Hmm, existing test does add an alias, the legacyMap.get() would returns null upon 'aliasAlg' (type: String). To test this particular scenario, I enhanced existing regression test to re-assign an alias and then check if the returned impl is as expected. The proposed test patch is attached. If we are including this test patch into this PR, then good to add the bug id to the test too.
[patch.txt](https://github.com/openjdk/jdk18/files/7771389/patch.txt)

shipilev: Thanks, @valeriepeng! Do you see the current JDK 18 failing with that test addition? Because this still passes for me without the fix from this PR: `CONF=linux-x86_64-server-fastdebug make run-test TEST=java/security/Provider/CaseSensitiveServices.java`

shipilev: Ping @valeriepeng ^^^

valeriepeng: The test update checks that the alias maps to the right algorithm. However, the particular code change covered by this fix is for removing the alias from the previous algorithm which is not triggered by the test change. Alias is an internal property of the Service object. Let me explore more and see if it can be triggered.

valeriepeng: There is no public API for retrieving the aliases from a given Service object. However, it's reported as part of toString(). So, we can check the toString() output to see if the alias is removed from the previous service. Updated test patch attached.
[patch.txt](https://github.com/openjdk/jdk18/files/7804357/patch.txt)

shipilev: All right, thank you, @valeriepeng! The new test fails without the patch, and passes with it. I added it to PR.

AlanBateman (approved):
Thanks for coordinating on this to ensure the change has a test.

valeriepeng (approved):

shipilev:
Thanks for reviews!

/integrate

PR #72, opened 2021-12-24
8279227: Access Bridge: Wrong frame position and hit test result on HiDPI display

VeselovAlex:
Perform scaling while converting sizes and coordinates to provide correct positions of accessible elements on HiDPI screen and make hit-test work properly. It uses FontRenderContext to receive display scale without API changes.

---------
### Progress

### Issue
 * [JDK-8279227](https://bugs.openjdk.java.net/browse/JDK-8279227): Access Bridge: Wrong frame position and hit test result on HiDPI display

### Reviewers
 * [Anton Tarasov](https://openjdk.java.net/census#ant) (@forantar - **Reviewer**) ⚠️ Review applies to 4ad508fee6fcdb84ea95223aba57d629cc0e4a06
 * [Alexander Zuev](https://openjdk.java.net/census#kizune) (@azuev-java - **Reviewer**)
 * [Alexey Ivanov](https://openjdk.java.net/census#aivanov) (@aivanov-jdk - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/72/head:pull/72` \
`$ git checkout pull/72`

`$ git checkout pull/72` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/72/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 72`

`$ git pr show -t 72`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/72.diff">https://git.openjdk.java.net/jdk18/pull/72.diff</a>

</details>

### DIFF
--- a/src/jdk.accessibility/windows/classes/com/sun/java/accessibility/internal/AccessBridge.java
+++ b/src/jdk.accessibility/windows/classes/com/sun/java/accessibility/internal/AccessBridge.java
@@ -479,19 +479,11 @@ public final class AccessBridge {
         if (parent == null) {
             return null;
         }
-        Point location = new Point(x, y);
-        InvocationUtils.invokeAndWait(new Callable<Void>() {
-            @Override
-            public Void call() throws Exception {
-                AffineTransform at = getTransformFromContext(parent);
-                if (at != null) {
-                    location.setLocation((int) (location.x / at.getScaleX()), (int) (location.y / at.getScaleY()));
-                }
-                return null;
-            }
-        }, parent);
-        x = location.x;
-        y = location.y;
+        GraphicsConfiguration gc = GraphicsEnvironment.getLocalGraphicsEnvironment()
+                .getDefaultScreenDevice().getDefaultConfiguration();
+        gc = AccessibilityGraphicsEnvironment.getGraphicsConfigurationAtPoint(gc, x, y);
+        x = AccessibilityGraphicsEnvironment.toUserSpaceX(gc, x);
+        y = AccessibilityGraphicsEnvironment.toUserSpaceY(gc, y);
         if (windowHandleToContextMap != null &&
             windowHandleToContextMap.containsValue(getRootAccessibleContext(parent))) {
             // Path for applications that register their top-level
@@ -1608,13 +1600,7 @@ public final class AccessBridge {
                                 r.x = p.x;
                                 r.y = p.y;
 
-                                AffineTransform at = getTransformFromContext(ac);
-                                if (at != null) {
-                                    r.x = (int) Math.floor(r.x * at.getScaleX());
-                                    r.y = (int) Math.floor(r.y * at.getScaleY());
-                                    r.width = (int) Math.ceil(r.width * at.getScaleX());
-                                    r.height = (int) Math.ceil(r.height * at.getScaleY());
-                                }
+                                r = AccessibilityGraphicsEnvironment.toDeviceSpaceAbs(r);
                                 return r;
                             }
                         } catch (Exception e) {
@@ -1757,20 +1743,6 @@ public final class AccessBridge {
         }, ac);
     }
 
-    /**
-     * return the screen coordinates transform from an AccessibleContext
-     */
-    private AffineTransform getTransformFromContext(final AccessibleContext ac) {
-        AccessibleComponent acomp = ac.getAccessibleComponent();
-        if (acomp != null) {
-            FontMetrics fm = acomp.getFontMetrics(acomp.getFont());
-            if (fm != null) {
-                return fm.getFontRenderContext().getTransform();
-            }
-        }
-
-        return null;
-    }
     /* ===== AccessibleText methods ===== */
 
     /**
@@ -2289,17 +2261,11 @@ public final class AccessBridge {
                     if (at != null) {
                         Rectangle rect = at.getCharacterBounds(index);
                         if (rect != null) {
-                            AffineTransform transform = getTransformFromContext(ac);
-                            if (transform != null) {
-                                rect.x = (int) Math.floor(rect.x * transform.getScaleX());
-                                rect.y = (int) Math.floor(rect.y * transform.getScaleY());
-                                rect.width = (int) Math.ceil(rect.width * transform.getScaleX());
-                                rect.height = (int) Math.ceil(rect.height * transform.getScaleY());
-                            }
                             String s = at.getAtIndex(AccessibleText.CHARACTER, index);
                             if (s != null && s.equals("\n")) {
                                 rect.width = 0;
                             }
+                            rect = AccessibilityGraphicsEnvironment.toDeviceSpaceAbs(rect);
                             return rect;
                         }
                     }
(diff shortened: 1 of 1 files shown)


forantar (approved):

aivanov-jdk:
@mrserb take a look

src/jdk.accessibility/windows/classes/com/sun/java/accessibility/internal/AccessBridge.java
```diff
@@ -1593,6 +1607,14 @@ public Rectangle call() throws Exception {
...
                                 r.x = p.x;
                                 r.y = p.y;
+
+                                AffineTransform at = getTransformFromContext(ac);
+                                if (at != null) {
+                                    r.x = (int) Math.floor(r.x * at.getScaleX());
+                                    r.y = (int) Math.floor(r.y * at.getScaleY());
+                                    r.width = (int) Math.ceil(r.width * at.getScaleX());
+                                    r.height = (int) Math.ceil(r.height * at.getScaleY());
+                                }
```
mrserb: This logic will not work on the multiscreen configs when the monitors have different DPI, the logic should be similar to the SunGraphicsEnvironment.toDeviceSpace().
and probably the round logic should use Region.clipRound as well.

src/jdk.accessibility/windows/classes/com/sun/java/accessibility/internal/AccessBridge.java
```diff
@@ -1735,6 +1757,20 @@ public AccessibleValue call() throws Exception {
...
 
+    /**
+     * return the screen coordinates transform from an AccessibleContext
+     */
+    private AffineTransform getTransformFromContext(final AccessibleContext ac) {
+        AccessibleComponent acomp = ac.getAccessibleComponent();
+        if (acomp != null) {
+            FontMetrics fm = acomp.getFontMetrics(acomp.getFont());
+            if (fm != null) {
+                return fm.getFontRenderContext().getTransform();
```
mrserb: I am not sure do we really update the FRC for the component when the DPI is changed?

src/jdk.accessibility/windows/classes/com/sun/java/accessibility/internal/AccessBridge.java
mrserb: If this data will be used in native then we can try to use device->ScaleDownAbsX/Y, etc methods.

VeselovAlex:
Hi, @mrserb, thanks for the review. Your idea is awesome and I've tried to implement it. It contains some copied code from `sun.java2d` but I can't import necessary methods without adding it to dependencies. 

But the issue with different scale factors on multi-display configuration is still present, frame size scaling seems to be stuck at primary display's scale factor. `jaccesibilityinspector` shows correctly scaled bounds, may it be an issue with NVDA?

src/jdk.accessibility/windows/classes/com/sun/java/accessibility/internal/AccessBridge.java
```diff
@@ -7338,4 +7347,121 @@ T getResult() throws Exception {
...
+     */
+    private static abstract class AccessibilityGraphicsEnvironment extends GraphicsEnvironment {
+        /**
+         * Returns the graphics configuration which bounds contain the given point.
+         *
+         * See {@link sun.java2d.SunGraphicsEnvironment#getGraphicsConfigurationAtPoint(GraphicsConfiguration, double, double)}
+         *
+         * @param  current the default configuration which is checked in the first
+         *         place
+         * @param  x the x coordinate of the given point
```
forantar: Please specify in which space the coordinates are expected.

src/jdk.accessibility/windows/classes/com/sun/java/accessibility/internal/AccessBridge.java
```diff
@@ -7338,4 +7347,121 @@ T getResult() throws Exception {
...
+         * @param  y the y coordinate of the given point
+         * @return the graphics configuration
+         */
+        public static GraphicsConfiguration getGraphicsConfigurationAtPoint(
+                GraphicsConfiguration current, double x, double y) {
+            if (current.getBounds().contains(x, y)) {
```
forantar: Here you do not translate the bounds to device space. Actually, I'd remove this check at all, as below you duplicate it. Also, when you call the method you pass the default GC, so it seems it does not bring much profit...

src/jdk.accessibility/windows/classes/com/sun/java/accessibility/internal/AccessBridge.java
forantar: I've checked it - it does not change, which is unexpected... Doesn't it seem like a bug?

Actually, the problem is that AccessibleComponent does not provide the component's GC, which is a missing. Ideally would be to add it to the API and then use - because it already contains actual scales for the component. It's a pity to do all that calculations just because we lack API calls...

Is it worth trying to fix the font metrics update instead and use it until the new API is ready and available?

mrserb: > I've checked it - it does not change, which is unexpected... Doesn't it seem like a bug?

Yes, I think this is a bug.

> Actually, the problem is that AccessibleComponent does not provide the component's GC, which is a missing. Ideally would be to add it to the API and then use - because it already contains actual scales for the component. It's a pity to do all that calculations just because we lack API calls...

But do we actually need the GC where the component is located, or do we need a GC where the a11y frame should be placed? I am not sure that both are always the same. Note that you can find the screen where the "virtual coordinates are located" and then convert them to the device space.

src/jdk.accessibility/windows/classes/com/sun/java/accessibility/internal/AccessBridge.java
mrserb: > Hi, @mrserb, thanks for the review. Your idea is awesome and I've tried to implement it. It contains some copied code from `sun.java2d` but I can't import necessary methods without adding it to dependencies.

Probably we can export that to the "jdk.accessibility" module? We already export "sun.awt" the same way. @prrace any thoughts?
But it will be good to make sure that it will work on a multiscreen config before doing that. At least need to understand the root cause of that.

src/jdk.accessibility/windows/classes/com/sun/java/accessibility/internal/AccessBridge.java
forantar: > But do we actually need the GC where the component is located, or do we need a GC where the a11y frame should be placed? I am not sure that both are always the same. Note that you can find the screen where the "virtual coordinates are located" and then convert them to the device space.

Well, in `AccessBridge.getAccessibleContextAt_1(int x, int y, AccessibleContext parent)` we expect [x, y] within the parent bounds and so we can assume [x, y] has the same scaling. However, `AccessBridge.getAccessibleContextAt_2` works with generic [x, y] and there we have to perform generic transformation, indeed...

Ok, if we can't avoid it, then I'm ok with the latest version (taking into account my inline comments). Thanks!

VeselovAlex:
I've fixed receiving GC at given point and now hit-test and frame bounds seem to be calculated correctly. But it looks like NVDA does not work correctly with different DPI on the second display, so it (at least v. 2021.3.1) has to be started in compatibility mode, see the attached picture

image

forantar:
Tested your patch with NVDA 2021.3.1 and it worked well for me.

forantar:
> I've fixed receiving GC at given point and now hit-test and frame bounds seem to be calculated correctly. But it looks like NVDA does not work correctly with different DPI on the second display, so it (at least v. 2021.3.1) has to be started in compatibility mode, see the attached picture
> 
Seems worth reporting this to NVDA (https://github.com/nvaccess/nvda/#communication-channels).

azuev-java (approved):
Tested on Windows 10 and Windows 11 with dual monitor configuration with different scale factors - last version of the fix seems to be working fine.

aivanov-jdk (approved):

forantar:
Based on 3 approves and successful testing this fix can be integrated (we are approaching RDP2 for JDK18).

PR #82, opened 2022-01-05
8279527: Dereferencing segments backed by different scopes leads to pollution

mcimadamore:
This patch fixes a performance issue when dereferencing memory segments backed by different kinds of scopes. When looking at inline traces, I found that one of our benchmark, namely `LoopOverPollutedSegment` was already hitting the ceiling of the bimorphic inline cache, specifically when checking liveness of the segment scope in the memory access hotpath (`ResourceScopeImpl::checkValidState`). The benchmark only used segments backed by confined and global scope. I then added (in the initialization "polluting" loop) segments backed by a shared scope, and then the benchmark numbers started to look as follows:

```
Benchmark                                              Mode  Cnt  Score   Error  Units
LoopOverPollutedSegments.heap_segment_floats_VH        avgt   30  7.004 ? 0.089  ms/op
LoopOverPollutedSegments.heap_segment_floats_instance  avgt   30  7.159 ? 0.016  ms/op
LoopOverPollutedSegments.heap_segment_ints_VH          avgt   30  7.017 ? 0.110  ms/op
LoopOverPollutedSegments.heap_segment_ints_instance    avgt   30  7.175 ? 0.048  ms/op
LoopOverPollutedSegments.heap_unsafe                   avgt   30  0.243 ? 0.004  ms/op
LoopOverPollutedSegments.native_segment_VH             avgt   30  7.366 ? 0.036  ms/op
LoopOverPollutedSegments.native_segment_instance       avgt   30  7.305 ? 0.098  ms/op
LoopOverPollutedSegments.native_unsafe                 avgt   30  0.238 ? 0.002  ms/op
```

That is, since now we have *three* different kinds of scopes (confined, shared and global), the call to the liveness check can no longer be inlined. One solution could be, as we do for the *base* accessor, to add a scope accessor to all memory segment implementation classes. But doing so only works ok for heap segments (for which the scope accessor just returns the global scope constants). For native segments, we're still megamorphic (as a native segment can be backed by all kinds of scopes).

In the end, it turned out to be much simpler to just make the liveness check monomorphic, since there's so much sharing between the code paths already. With that change, numbers of the tweaked benchmark go back to normal:

```
Benchmark                                              Mode  Cnt  Score   Error  Units
LoopOverPollutedSegments.heap_segment_floats_VH        avgt   30  0.241 ? 0.003  ms/op
LoopOverPollutedSegments.heap_segment_floats_instance  avgt   30  0.244 ? 0.003  ms/op
LoopOverPollutedSegments.heap_segment_ints_VH          avgt   30  0.242 ? 0.003  ms/op
LoopOverPollutedSegments.heap_segment_ints_instance    avgt   30  0.248 ? 0.001  ms/op
LoopOverPollutedSegments.heap_unsafe                   avgt   30  0.247 ? 0.013  ms/op
LoopOverPollutedSegments.native_segment_VH             avgt   30  0.245 ? 0.004  ms/op
LoopOverPollutedSegments.native_segment_instance       avgt   30  0.245 ? 0.001  ms/op
LoopOverPollutedSegments.native_unsafe                 avgt   30  0.247 ? 0.005  ms/op
```

Note that this patch tidies up a bit the usage of `checkValidState` vs. `checkValidStateSlow`. The former should only really be used in the hot path, while the latter is a more general routine which should be used in non-performance critical code. Making `checkValidState` monomorphic caused the `ScopeAccessError` to be generated in more places, so I needed to either update the usage to use the safer `checkValidStateSlow` (where possible) or, (in `Buffer` and `ConfinedScope`) just add extra wrapping.

---------
### Progress

### Issue
 * [JDK-8279527](https://bugs.openjdk.java.net/browse/JDK-8279527): Dereferencing segments backed by different scopes leads to pollution

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - **Reviewer**) ⚠️ Review applies to 04a1e9f24485390d2db100ea1ce09049a73783f2
 * [Jorn Vernee](https://openjdk.java.net/census#jvernee) (@JornVernee - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/82/head:pull/82` \
`$ git checkout pull/82`

`$ git checkout pull/82` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/82/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 82`

`$ git pr show -t 82`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/82.diff">https://git.openjdk.java.net/jdk18/pull/82.diff</a>

</details>

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
@@ -364,11 +364,7 @@ public abstract non-sealed class AbstractMemorySegmentImpl extends MemorySegment
     }
 
     void checkValidState() {
-        try {
-            scope.checkValidState();
-        } catch (ScopedMemoryAccess.Scope.ScopedAccessError ex) {
-            throw new IllegalStateException("This segment is already closed");
-        }
+        scope.checkValidStateSlow();
     }
 
     @Override
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ConfinedScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ConfinedScope.java
@@ -25,7 +25,6 @@
 
 package jdk.internal.foreign;
 
-import jdk.internal.misc.ScopedMemoryAccess;
 import jdk.internal.vm.annotation.ForceInline;
 
 import java.lang.invoke.MethodHandles;
@@ -40,9 +39,7 @@ import java.lang.ref.Cleaner;
  */
 final class ConfinedScope extends ResourceScopeImpl {
 
-    private int lockCount = 0;
     private int asyncReleaseCount = 0;
-    private final Thread owner;
 
     static final VarHandle ASYNC_RELEASE_COUNT;
 
@@ -56,7 +53,6 @@ final class ConfinedScope extends ResourceScopeImpl {
 
     public ConfinedScope(Thread owner, Cleaner cleaner) {
         super(owner, new ConfinedResourceList(), cleaner);
-        this.owner = owner;
     }
 
     @Override
@@ -68,17 +64,17 @@ final class ConfinedScope extends ResourceScopeImpl {
     @ForceInline
     public void acquire0() {
         checkValidStateSlow();
-        if (lockCount == MAX_FORKS) {
+        if (state == MAX_FORKS) {
             throw new IllegalStateException("Scope keep alive limit exceeded");
         }
-        lockCount++;
+        state++;
     }
 
     @Override
     @ForceInline
     public void release0() {
         if (Thread.currentThread() == owner) {
-            lockCount--;
+            state--;
         } else {
             // It is possible to end up here in two cases: this scope was kept alive by some other confined scope
             // which is implicitly released (in which case the release call comes from the cleaner thread). Or,
@@ -90,10 +86,10 @@ final class ConfinedScope extends ResourceScopeImpl {
 
     void justClose() {
         checkValidStateSlow();
-        if (lockCount == 0 || lockCount - ((int)ASYNC_RELEASE_COUNT.getVolatile(this)) == 0) {
+        if (state == 0 || state - ((int)ASYNC_RELEASE_COUNT.getVolatile(this)) == 0) {
             state = CLOSED;
         } else {
-            throw new IllegalStateException("Scope is kept alive by " + lockCount + " scopes");
+            throw new IllegalStateException("Scope is kept alive by " + state + " scopes");
         }
     }
 
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ConfinedScope.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ConfinedScope.java
@@ -25,6 +25,7 @@
 
 package jdk.internal.foreign;
 
+import jdk.internal.misc.ScopedMemoryAccess;
 import jdk.internal.vm.annotation.ForceInline;
 
 import java.lang.invoke.MethodHandles;
@@ -39,7 +40,6 @@ import java.lang.ref.Cleaner;
  */
 final class ConfinedScope extends ResourceScopeImpl {
 
-    private boolean closed; // = false
     private int lockCount = 0;
     private int asyncReleaseCount = 0;
     private final Thread owner;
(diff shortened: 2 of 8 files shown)


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/AbstractMemorySegmentImpl.java
```diff
@@ -364,11 +364,7 @@ public void checkAccess(long offset, long length, boolean readOnly) {
     }
 
     void checkValidState() {
-        try {
-            scope.checkValidState();
-        } catch (ScopedMemoryAccess.Scope.ScopedAccessError ex) {
-            throw new IllegalStateException("This segment is already closed");
-        }
+        scope.checkValidStateSlow();
```
mcimadamore, line 367: Not to be confused with `ResourceScope::checkValidState` - this method is only really used by other non-performance critical code around `AbstractMemorySegmentImpl`.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ResourceScopeImpl.java
```diff
@@ -147,22 +167,33 @@ public void close() {
...
+     * relies on invariants associated with the memory scope implementations (volatile access
+     * to the closed state bit is replaced with plain access). This method should be monomorphic,
+     * to avoid virtual calls in the memory access hot path. This method is not intended as general purpose method
+     * and should only be used in the memory access handle hot path; for liveness checks triggered by other API methods,
+     * please use {@link #checkValidStateSlow()}.
      */
-    public abstract void checkValidState();
+    @ForceInline
+    public final void checkValidState() {
+        if (owner != null && owner != Thread.currentThread()) {
```
PaulSandoz, line 190: For consistency we could change code `checkValidStateSlow` to refer directly to `owner`.

It would be satisfying, but I don't know if it's possible, to compose `checkValidStateSlow` from `checkValidState` e.g.
```java
public final checkValidStateSlow() {
    checkValidState();
    if (!isAlive() { ... }
}
```

mcimadamore: I'll change use of `owner`. It's not really possible to write checkValidStateSlow in terms of checkValidState, because the latter does a plain read of the state, whereas the former does a volatile read. Reusing one from the other would result in two reads (a plain and a volatile).

PaulSandoz: Ok. My thought was that since this is slow two reads do not matter, but i did not reason fully about the concurrent implications (if the fast alive check returns false, the slow alive check can still return true so that seems good, if the fast check returns true i was presume the slow alive check would also be true, given the way state changes monotonically?)

PaulSandoz (approved):

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ResourceScopeImpl.java
mcimadamore: If we're ok with a redundant plain read, then I don't think there are issues. You just do two reads, and the latter (the volatile one) is the one that counts. I don't think we can rely much on dependencies between what the plain read and what the volatile read will see. The state is updated in both direction (for shared segments) e.g. we can go from ALIVE to CLOSING then back to ALIVE. Or we could go from ALIVE to CLOSING to CLOSE.

That said, I guess my main reservation for writing one routine on top of the other is that we really want checkValidState to be only used in critical hot paths. It has a non-volatile semantics and an exception handling which only really makes sense when combined with ScopedMemoryAccess - for this reason, using it as an internal building primitive didn't seem to me as a great idea.

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ConfinedScope.java
```diff
@@ -39,7 +40,6 @@
  */
 final class ConfinedScope extends ResourceScopeImpl {
 
-    private boolean closed; // = false
     private int lockCount = 0;
     private int asyncReleaseCount = 0;
     private final Thread owner;
```
JornVernee: Is this `owner` field still needed now that the super class also has it?

JornVernee (approved):

src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/ConfinedScope.java
mcimadamore: Forgot to remove - well spotted

JornVernee (approved):

PR #88, opened 2022-01-10
8279356: Method linking fails with guarantee(mh->adapter() != NULL) failed: Adapter blob must already exist!

TobiHartmann:
Adapter creation during method linking may fail due to a lack of code cache space which leads to a `VirtualMachineError` being thrown and thus a bail out from linking the holder klass:
https://github.com/openjdk/jdk18/blob/d65c665839c0a564c422ef685f2673fac37315d7/src/hotspot/share/oops/method.cpp#L1239-L1247

If the `VirtualMachineError` is handled/ignored by the application, we may later attempt to link the same klass and therefore also the same method again. We then incorrectly bail out from adapter creation because the `_i2i_entry` is set, assuming that this can only happen if adapters have already been created. However, that is not guaranteed because the interpreter entry is set right **before** adapters are created:
https://github.com/openjdk/jdk18/blob/d65c665839c0a564c422ef685f2673fac37315d7/src/hotspot/share/oops/method.cpp#L1213-L1230

I propose to instead check if adapters have been created.

This is an old bug that was just recently triggered by an unrelated change.

Thanks,
Tobias

---------
### Progress

### Issue
 * [JDK-8279356](https://bugs.openjdk.java.net/browse/JDK-8279356): Method linking fails with guarantee(mh->adapter() != NULL) failed: Adapter blob must already exist!

### Reviewers
 * [Christian Hagedorn](https://openjdk.java.net/census#chagedorn) (@chhagedorn - **Reviewer**) ⚠️ Review applies to e0ba3cfb4705af43aaffc553295786bda711d833
 * [Vladimir Kozlov](https://openjdk.java.net/census#kvn) (@vnkozlov - **Reviewer**) ⚠️ Review applies to e0ba3cfb4705af43aaffc553295786bda711d833
 * [Coleen Phillimore](https://openjdk.java.net/census#coleenp) (@coleenp - **Reviewer**) ⚠️ Review applies to e0ba3cfb4705af43aaffc553295786bda711d833

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/88/head:pull/88` \
`$ git checkout pull/88`

`$ git checkout pull/88` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/88/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 88`

`$ git pr show -t 88`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/88.diff">https://git.openjdk.java.net/jdk18/pull/88.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/oops/method.cpp
+++ b/src/hotspot/share/oops/method.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/src/hotspot/share/oops/method.cpp
+++ b/src/hotspot/share/oops/method.cpp
@@ -1198,7 +1198,7 @@ void Method::unlink_method() {
 void Method::link_method(const methodHandle& h_method, TRAPS) {
   // If the code cache is full, we may reenter this function for the
   // leftover methods that weren't linked.
-  if (_i2i_entry != NULL) {
+  if (adapter() != NULL) {
     return;
   }
   assert( _code == NULL, "nothing compiled yet" );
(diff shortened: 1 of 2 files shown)


chhagedorn (approved):
That looks reasonable!

TobiHartmann:
Thanks, Christian!

vnkozlov (approved):
Good.

coleenp (approved):
Looks good.  You should update the copyrights to 2022 though.

TobiHartmann:
Vladimir, Coleen, thanks for the reviews. I updated the copyright.

PR #93, opened 2022-01-11
8279597: [TESTBUG] ReturnBlobToWrongHeapTest.java fails with -XX:TieredStopAtLevel=1 on machines with many cores

shqking:
This test failure occurred on machines with a lot of cores. Here is a
snippet from the log file:

```
----------System.out:(14/1085)----------
CompileCommand: dontinline compiler/codecache/stress/Helper$TestCase.method bool dontinline = true
[0.046s][warning][codecache] CodeHeap 'non-profiled nmethods' is full. Compiler has been disabled.
[0.046s][warning][codecache] Try increasing the code heap size using -XX:NonProfiledCodeHeapSize=
CodeHeap 'non-profiled nmethods': size=8Kb used=7Kb max_used=7Kb free=0Kb
 bounds [0x0000ffff97b55000, 0x0000ffff97b57000, 0x0000ffff97b57000]
CodeHeap 'non-nmethods': size=16376Kb used=941Kb max_used=941Kb free=15434Kb
 bounds [0x0000ffff96b57000, 0x0000ffff96dc7000, 0x0000ffff97b55000]
 total_blobs=269 nmethods=8 adapters=194
 compilation: disabled (not enough contiguous free space left)
 stopped_count=1, restarted_count=0
 full_count=1
Error occurred during initialization of boot layer
java.lang.InternalError: java.lang.NoSuchMethodException: no such method: java.lang.invoke.MethodHandle.linkToStatic(MemberName)Object/invokeStatic
Caused by: java.lang.NoSuchMethodException: no such method: java.lang.invoke.MethodHandle.linkToStatic(MemberName)Object/invokeStatic
```

On machines with many cores, the number of C1 temporary code buffers is
big. See [1]. In my local test on a machine with 224 cores, `c1_count`
equals to 21, and the total size for non-nmethod heap increases to about
16.5M.

In this test case, `ReservedCodeCacheSize` is set as 16M, which is not
enough for the non-nmethod heap. See [2]. As a result, non-profiled heap
is set to (twice) the minimal size. As shown in the log file, the size
is only 8Kb, which is very small and leads to failure at initialization
phase.

In this patch, we increase `ReservedCodeCacheSize` from 16M to 64M,
which is big enough for machines with 1024 cores. Note that we continue
to use the initial value for variable `largeBlobSize`.

[1] https://github.com/openjdk/jdk/blob/master/src/hotspot/share/code/codeCache.cpp#L200
[2] https://github.com/openjdk/jdk/blob/master/src/hotspot/share/code/codeCache.cpp#L216

---------
### Progress

### Issue
 * [JDK-8279597](https://bugs.openjdk.java.net/browse/JDK-8279597): [TESTBUG] ReturnBlobToWrongHeapTest.java fails with -XX:TieredStopAtLevel=1 on machines with many cores

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/93/head:pull/93` \
`$ git checkout pull/93`

`$ git checkout pull/93` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/93/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 93`

`$ git pr show -t 93`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/93.diff">https://git.openjdk.java.net/jdk18/pull/93.diff</a>

</details>

### DIFF
--- a/test/hotspot/jtreg/compiler/codecache/stress/ReturnBlobToWrongHeapTest.java
+++ b/test/hotspot/jtreg/compiler/codecache/stress/ReturnBlobToWrongHeapTest.java
@@ -35,8 +35,9 @@
  *                   -XX:+WhiteBoxAPI
  *                   -XX:CompileCommand=dontinline,compiler.codecache.stress.Helper$TestCase::method
  *                   -XX:+SegmentedCodeCache
- *                   -XX:ReservedCodeCacheSize=64M
+ *                   -XX:ReservedCodeCacheSize=16M
  *                   -XX:CodeCacheMinBlockLength=1
+ *                   -XX:CICompilerCount=2
  *                   compiler.codecache.stress.ReturnBlobToWrongHeapTest
  */
 
@@ -47,7 +48,7 @@ import sun.hotspot.code.BlobType;
 import java.util.ArrayList;
 
 public class ReturnBlobToWrongHeapTest {
-    private static final long largeBlobSize = Helper.WHITE_BOX.getUintxVMFlag("ReservedCodeCacheSize") >> 8;
+    private static final long largeBlobSize = Helper.WHITE_BOX.getUintxVMFlag("ReservedCodeCacheSize") >> 6;
     private static final long codeCacheMinBlockLength = Helper.WHITE_BOX.getUintxVMFlag("CodeCacheMinBlockLength");
     private static final BlobType[] BLOB_TYPES = BlobType.getAvailable().toArray(new BlobType[0]);
 
--- a/test/hotspot/jtreg/compiler/codecache/stress/ReturnBlobToWrongHeapTest.java
+++ b/test/hotspot/jtreg/compiler/codecache/stress/ReturnBlobToWrongHeapTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -35,7 +35,7 @@
  *                   -XX:+WhiteBoxAPI
  *                   -XX:CompileCommand=dontinline,compiler.codecache.stress.Helper$TestCase::method
  *                   -XX:+SegmentedCodeCache
- *                   -XX:ReservedCodeCacheSize=16M
+ *                   -XX:ReservedCodeCacheSize=64M
  *                   -XX:CodeCacheMinBlockLength=1
  *                   compiler.codecache.stress.ReturnBlobToWrongHeapTest
  */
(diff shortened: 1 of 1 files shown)


vnkozlov (commented):
There is specific code to take into account CodeCache size when we calculate number of compiler thread:
https://github.com/openjdk/jdk18/blob/master/src/hotspot/share/compiler/compilationPolicy.cpp#L450

Can you check why it does not work?

There is also possibility that number of compiler threads increase dynamically which is guarded by `-XX:+UseDynamicNumberOfCompilerThreads` flag. Try to switch it off in this test.

vnkozlov:
You can also set small value with `-XX:CICompilerCount=n' for this test. Increasing CodeCachecsize to 64M may help your case but we can have machines with core count even larger.

shqking:
> There is specific code to take into account CodeCache size when we calculate number of compiler thread: https://github.com/openjdk/jdk18/blob/master/src/hotspot/share/compiler/compilationPolicy.cpp#L450
> 
> Can you check why it does not work?

Thanks a lot for your review.

It doesn't work because the check at https://github.com/openjdk/jdk18/blob/master/src/hotspot/share/compiler/compilationPolicy.cpp#L461 fails.
In my local test machine(with 224 cores), `count` equals to 21, but `max_count` equals to 28 (Note that `ReservedCodeCacheSize` is 16M, `CodeCacheMinimumUseSpace` is 400K and `buffer_size` is 576,716).


> 
> There is also possibility that number of compiler threads increase dynamically which is guarded by `-XX:+UseDynamicNumberOfCompilerThreads` flag. Try to switch it off in this test.

I tried to switch it off, but the test still failed.

shqking:
> You can also set small value with `-XX:CICompilerCount=n' for this test. Increasing CodeCachecsize to 64M may help your case but we can have machines with core count even larger.

Yes. Agree. I think your solution would be better.
Will test more and update soon. Thanks.

shqking:
> > You can also set small value with `-XX:CICompilerCount=n' for this test. Increasing CodeCachecsize to 64M may help your case but we can have machines with core count even larger.
> 
> Yes. Agree. I think your solution would be better. Will test more and update soon. Thanks.

@vnkozlov I'm afraid it doesn't work either. 

If we set CICompilerCount to a small value, e.g., 2, the test case
1) passed on server/client builds as expected
2) but **failed** on non-compiler JVM build, e.g., zero, with the following error msg:  `CICompilerCount (2) cannot be greater than 0 because there are no compilers`

Because there are constraints for the valid values of CICompilerCount. See https://github.com/openjdk/jdk18/blob/master/src/hotspot/share/runtime/flags/jvmFlagConstraintsCompiler.cpp#L54

shqking:
> > > You can also set small value with `-XX:CICompilerCount=n' for this test. Increasing CodeCachecsize to 64M may help your case but we can have machines with core count even larger.
> > 
> > 
> > Yes. Agree. I think your solution would be better. Will test more and update soon. Thanks.
> 
> @vnkozlov I'm afraid it doesn't work either.
> 
> If we set CICompilerCount to a small value, e.g., 2, the test case
> 
> 1. passed on server/client builds as expected
> 2. but **failed** on non-compiler JVM build, e.g., zero, with the following error msg:  `CICompilerCount (2) cannot be greater than 0 because there are no compilers`
> 
> Because there are constraints for the valid values of CICompilerCount. See https://github.com/openjdk/jdk18/blob/master/src/hotspot/share/runtime/flags/jvmFlagConstraintsCompiler.cpp#L54

On a second thought, I suppose it might be okay for this test case to **FAIL** with **ZERO** JVM build.

1) this test case is located under `test/hotspot/jtreg/compiler/*`, and I think those test cases are designed to test JVM **WITH compilers**.
2) there exist a number of test cases where CICompilerCount is specified, e.g.,
  `test/hotspot/jtreg/compiler/loopopts/superword/TestNegBaseOffset.java
  test/hotspot/jtreg/compiler/classUnloading/methodUnloading/TestOverloadCompileQueues.java`
  And in my local test, these test cases **failed as expected** with ZERO JVM build.

Hence, my point is that we could follow your solution, i.e. setting `CICompilerCout` to a small value. @vnkozlov 
Please check the new commit.

vnkozlov (approved):
Good  and trivial.

shqking:
Thanks for your review! I think this trivial patch is ready to be merged.

/integrate

PR #97, opened 2022-01-12
8279930: Synthetic cast causes generation of store barriers when using heap segments

mcimadamore:
When looking at larger benchmarks, I noted a discrepancy between the performance of off-heap segments and on-heap segments. Looking at the assembly for the `MemorySegment::asSlice` method I could see some additional barriers in the off-heap case, but could not initially make sense of them. Vlad pointed me at G1 (in fact no such barrier was emitted when using a different GC, such as the serial GC, or ZGC), and later Erik narrowed the problem down to a failure in a C2 optimization to remove barriers around initializing stores. This problem was caused by a synthetic cast added by javac to a value (the base object) that initialized the newly created memory segment slice. Because of that case, C2 missed the store as an "initializing" one, and inserted additional barriers. This patch should make performance of on-heap segments a lot more reliable, especially when slicing is involved.

---------
### Progress

### Issue
 * [JDK-8279930](https://bugs.openjdk.java.net/browse/JDK-8279930): Synthetic cast causes generation of store barriers when using heap segments

### Reviewers
 * [Paul Sandoz](https://openjdk.java.net/census#psandoz) (@PaulSandoz - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/97/head:pull/97` \
`$ git checkout pull/97`

`$ git checkout pull/97` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/97/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 97`

`$ git pr show -t 97`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/97.diff">https://git.openjdk.java.net/jdk18/pull/97.diff</a>

</details>

### DIFF
--- a/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/HeapMemorySegmentImpl.java
+++ b/src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/HeapMemorySegmentImpl.java
@@ -42,7 +42,7 @@ import java.util.Objects;
  * of {@link HeapMemorySegmentImpl} are defined (e.g. {@link OfFloat}, so that each subclass can override the
  * {@link HeapMemorySegmentImpl#base()} method so that it returns an array of the correct (sharp) type.
  */
-public abstract class HeapMemorySegmentImpl<H> extends AbstractMemorySegmentImpl {
+public abstract class HeapMemorySegmentImpl extends AbstractMemorySegmentImpl {
 
     private static final Unsafe UNSAFE = Unsafe.getUnsafe();
     private static final int BYTE_ARR_BASE = UNSAFE.arrayBaseOffset(byte[].class);
@@ -53,17 +53,17 @@ public abstract class HeapMemorySegmentImpl<H> extends AbstractMemorySegmentImpl
     private static final long MAX_ALIGN_8 = 8;
 
     final long offset;
-    final H base;
+    final Object base;
 
     @ForceInline
-    HeapMemorySegmentImpl(long offset, H base, long length, int mask) {
+    HeapMemorySegmentImpl(long offset, Object base, long length, int mask) {
         super(length, mask, ResourceScopeImpl.GLOBAL);
         this.offset = offset;
         this.base = base;
     }
 
     @Override
-    abstract H base();
+    abstract Object base();
 
     @Override
     long min() {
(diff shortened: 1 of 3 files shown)


src/jdk.incubator.foreign/share/classes/jdk/internal/foreign/HeapMemorySegmentImpl.java
```diff
@@ -84,9 +84,9 @@ ByteBuffer makeByteBuffer() {
 
     // factories
 
-    public static class OfByte extends HeapMemorySegmentImpl<byte[]> {
+    public static class OfByte extends HeapMemorySegmentImpl {
 
-        OfByte(long offset, byte[] base, long length, int mask) {
+        OfByte(long offset, Object base, long length, int mask) {
```
PaulSandoz, line 93: Is it possible to retain the array type for the constructor?

mcimadamore: yes, that should be harmless

mcimadamore: Actually, no, sorry. If I sharpen the type, then `dup` has to cast, so we go back to the original issue.

PaulSandoz: Ah yes, i see. Maybe it's worth adding a general comment to the class doc of `HeapMemorySegmentImpl` briefly explaining the reasons for use of `Object`.

test/micro/org/openjdk/bench/jdk/incubator/foreign/LoopOverSlice.java
```diff
@@ -0,0 +1,172 @@
...
+    static final int ALLOC_SIZE = ELEM_SIZE * CARRIER_SIZE;
+
+    MemorySegment nativeSegment, heapSegment;
+    ResourceScope scope;
+
+    @Setup
+    public void setup() {
+        scope = ResourceScope.newConfinedScope();
+        nativeSegment = MemorySegment.allocateNative(ALLOC_SIZE, scope);
+        heapSegment = MemorySegment.ofArray(new float[ELEM_SIZE]);
```
PaulSandoz: The heap `MemorySegment` wraps `float[]` but is accessed as if `int[]`, same bit size, but I suspect it was not intentional?

PaulSandoz (approved):

PaulSandoz (approved):

PR #104, opened 2022-01-16
8273383: vmTestbase/vm/gc/containers/Combination05/TestDescription.java crashes verifying length of DCQS

kimbarrett:
Please review this fix of an ABA problem in G1DirtyCardQueueSet's use of
NonblockingQueue. 

Consider queue q contains one entry, `X`.

Thread1 begins a push/append operation. WOLG, assume a push of `Y`. It
atomically exchanges `q._tail` with `Y`, retaining the old value of `q._tail`,
which is `X`. Pause Thread1 here.

Thread2 pops `X` from the queue. It successfully changes `X._next` from
end-marker to null and `q._head` from `X` to null. Thread2 fails to change
`q._tail` from `X` to null, as Thread1 already changed it to `Y`.

Various threads add entries to the queue after `Y`, and various threads try to
pop entries and find the queue "empty" because `q._head` will remain null
until Thread1 proceeds.

In parallel with entries being added, Thread2 processes `X` and releases it
for reuse. Some thread reallocates it, fills it, and pushes it onto the
queue. We now have a chain from `Y` => `X` in `q._tail` with `q._head` still
null.

If Thread1 resumes at that point, it finds `X._next` == end-marker,
successfully changes it to `Y`, and "completes". We now have `q._head` == null
and `q._tail` == `X`, with `X` => `Y` => `X`. The next push/append will
observe `q._tail` of `X` having a non-end-marker `_next` and change both
`q._head` and `q._tail`, dropping the list containing `X`, `Y`, &etc on the
floor.

There may be other similarly bad scenarios, depending on exactly what is going
on when Thread1 resumes.

All of these are fixed by performing the push operation in a critical section,
so that `X` cannot be reused until the push that is observing it
completes. That is, in `DirtyCardQueueSet::enqueue_completed_buffer`, put
`_completed.push(*cbn);` in a GlobalCounter critical section.

Testing:
mach5 tier1

Running the failing test with -XX:G1UpdateBufferSize=3 was failing at a rate
of about 1/1K runs of the test.  With this change, there were no failures in
more than 60K runs.

---------
### Progress

### Issue
 * [JDK-8273383](https://bugs.openjdk.java.net/browse/JDK-8273383): vmTestbase/vm/gc/containers/Combination05/TestDescription.java crashes verifying length of DCQS

### Reviewers
 * [Thomas Schatzl](https://openjdk.java.net/census#tschatzl) (@tschatzl - **Reviewer**) ⚠️ Review applies to 9fd316f5f39306dffee8dd33b8017b0ae4094558
 * [Stefan Johansson](https://openjdk.java.net/census#sjohanss) (@kstefanj - **Reviewer**) ⚠️ Review applies to 9fd316f5f39306dffee8dd33b8017b0ae4094558

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/104/head:pull/104` \
`$ git checkout pull/104`

`$ git checkout pull/104` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/104/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 104`

`$ git pr show -t 104`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/104.diff">https://git.openjdk.java.net/jdk18/pull/104.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/gc/g1/g1DirtyCardQueue.cpp
+++ b/src/hotspot/share/gc/g1/g1DirtyCardQueue.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2001, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2001, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -123,7 +123,13 @@ void G1DirtyCardQueueSet::enqueue_completed_buffer(BufferNode* cbn) {
   // Increment _num_cards before adding to queue, so queue removal doesn't
   // need to deal with _num_cards possibly going negative.
   size_t new_num_cards = Atomic::add(&_num_cards, buffer_size() - cbn->index());
-  _completed.push(*cbn);
+  {
+    // Perform push in CS.  The old tail may be popped while the push is
+    // observing it (attaching it to the new buffer).  We need to ensure it
+    // can't be reused until the push completes, to avoid ABA problems.
+    GlobalCounter::CriticalSection cs(Thread::current());
+    _completed.push(*cbn);
+  }
   if ((new_num_cards > process_cards_threshold()) &&
       (_primary_refinement_thread != NULL)) {
     _primary_refinement_thread->activate();


tschatzl (approved):
Lgtm.

kstefanj (approved):
Fix looks good. 

In addition I think we should update the docs in the `nonblockingQueue.hpp` to say that `append(...)`/`push(...)` are also subject to ABA behavior, like we do for `try_pop(...)`.

Also spent some time thinking about the `append(...)` done in `enqueue_paused_buffers_aux(...)`, can we guarantee that an element can't re-enter the queue this way under the special circumstances just after a safepoint where buffers are paused "wrongly" (if this can happen). I guess this might be prevented by, the fact that no other thread could process the element in the first place making it impossible for something to be "re-used" at this point.

kimbarrett:
> Fix looks good.
> 
> In addition I think we should update the docs in the `nonblockingQueue.hpp` to say that `append(...)`/`push(...)` are also subject to ABA behavior, like we do for `try_pop(...)`.

I was planning to do that as one of several planned JDK 19 RFEs, while keeping
the bug fix change minimal.  If you prefer, I can bring forward the comment
changes.

> Also spent some time thinking about the `append(...)` done in `enqueue_paused_buffers_aux(...)`, can we guarantee that an element can't re-enter the queue this way under the special circumstances just after a safepoint where buffers are paused "wrongly" (if this can happen). I guess this might be prevented by, the fact that no other thread could process the element in the first place making it impossible for something to be "re-used" at this point.

The buffers that get pushed by enqueue_previous_paused_buffers were taken from
the queue and paused before the previous safepoint. No thread can be in
push/append or try_pop (via DCQS) across a safepoint boundary. So no ABA.

enqueue_all_paused_buffers is called by the VMThread during a safepoint, with
no concurrent push/append or try_pop operations.  So again, no ABA.

kstefanj:
> I was planning to do that as one of several planned JDK 19 RFEs, while keeping the bug fix change minimal. If you prefer, I can bring forward the comment changes.
> 
I would prefer but if you plan bigger changes to the comments I'm fine with leaving it to 19.

> 
> The buffers that get pushed by enqueue_previous_paused_buffers were taken from the queue and paused before the previous safepoint. No thread can be in push/append or try_pop (via DCQS) across a safepoint boundary. So no ABA.
> 
I was mostly thinking about this case, that you mentioned. That for a short period of time after a safepoint some threads might still think they should yield. In that case it is not clear to me when such paused buffers would be re-added.

kimbarrett:
> > I was planning to do that as one of several planned JDK 19 RFEs, while keeping the bug fix change minimal. If you prefer, I can bring forward the comment changes.
> 
> I would prefer but if you plan bigger changes to the comments I'm fine with leaving it to 19.

I'm not sure; I haven't written the comment updates yet.  There might not be much that needs to be changed.  I'll look through all the changes and see which other ones (if any) need such updates.

> > The buffers that get pushed by enqueue_previous_paused_buffers were taken from the queue and paused before the previous safepoint. No thread can be in push/append or try_pop (via DCQS) across a safepoint boundary. So no ABA.
> 
> I was mostly thinking about this case, that you mentioned. That for a short period of time after a safepoint some threads might still think they should yield. In that case it is not clear to me when such paused buffers would be re-added.

That case you describe is https://bugs.openjdk.java.net/browse/JDK-8278475.  The buffers that get paused then won't be re-enqueued until after the next safepoint.

kstefanj:
> That case you describe is https://bugs.openjdk.java.net/browse/JDK-8278475. The buffers that get paused then won't be re-enqueued until after the next safepoint.

Thanks for clarifying this. Then I'm all good with the change 👍

kimbarrett:
Thanks @kstefanj and @tschatzl  for reviews.

PR #106, opened 2022-01-17
8279294: NonblockingQueue::try_pop may improperly indicate queue is empty

kimbarrett:
Please review this improvement to NonblockingQueue::try_pop.  The old code
returned an indication that the queue was empty in some cases where that
wasn't true.  In particular, contending try_pop operations could result in
some incorrectly indicating empty.  The change fixes that and improves the
interaction between contending try_pops.

Testing:
mach5 tier1

Lots of testing of this change in conjunction with others as part of
investigating and fixing JDK-8273383.

---------
### Progress
- [ ] Change must be properly reviewed

### Issue
 * [JDK-8279294](https://bugs.openjdk.java.net/browse/JDK-8279294): NonblockingQueue::try_pop may improperly indicate queue is empty

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jdk18 pull/106/head:pull/106` \
`$ git checkout pull/106`

`$ git checkout pull/106` \
`$ git pull https://git.openjdk.java.net/jdk18 pull/106/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 106`

`$ git pr show -t 106`

</details>

<a href="https://git.openjdk.java.net/jdk18/pull/106.diff">https://git.openjdk.java.net/jdk18/pull/106.diff</a>

</details>

### DIFF
--- a/src/hotspot/share/utilities/nonblockingQueue.inline.hpp
+++ b/src/hotspot/share/utilities/nonblockingQueue.inline.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


dholmes-ora:
Hi Kim,

I haven't looked at this implementation in detail before, but what you are describing as a bug seems to be a known "property" of the implementation:
```
// A queue may temporarily appear to be empty even though elements have been
// added and not removed.  For example, after running the following program,
// the value of r may be NULL.
//
// thread1: q.push(a); r = q.pop();
// thread2: q.push(b);
//
// This can occur if the push of b started before the push of a, but didn't
// complete until after the pop.
```

??

David

kimbarrett:
> Hi Kim,
> 
> I haven't looked at this implementation in detail before, but what you are describing as a bug seems to be a known "property" of the implementation:
> 
> ```
> // A queue may temporarily appear to be empty even though elements have been
> // added and not removed.  For example, after running the following program,
> // the value of r may be NULL.
> //
> // thread1: q.push(a); r = q.pop();
> // thread2: q.push(b);
> //
> // This can occur if the push of b started before the push of a, but didn't
> // complete until after the pop.
> ```
> 
> ??
> 
> David

That's a different situation, and seems unavoidable without a pretty different design.

The case we're trying to squash with this change can exhibit even without any
concurrent push/append, i.e.
1. phase1 collects entries in the queue.
2. phase change ensures phase1 is complete and phase2 not yet started.
3. phase2 takes entries from the queue.

During phase2, the bug being addressed could lead concurrent try_pops to have
some (possibly even all but one) threads conclude there are no more entries
and so no more work to do, even though there are actually lots of entries and
all those threads just lost the race for the next one.

That isn't necessary or intended behavior.

src/hotspot/share/utilities/nonblockingQueue.inline.hpp
```diff
@@ -135,28 +135,40 @@ bool NonblockingQueue<T, next_ptr>::try_pop(T** node_ptr) {
...
-    // lock-free pop from the head of a singly linked list to try to take it.
-    if (result == Atomic::cmpxchg(&_head, result, next_node)) {
-      // Former head successfully taken.
+  if (!is_end(next_node)) {
+    // [Clause 1]
+    // Attempt to advance the list, moving next_node to the head and
+    // shifting result out of the list.  There are several cases.
+    // (1) next_node is the extension of the queue's list.
+    // (2) next_node is NULL, because a competing try_pop took result.
+    // (3) next_node is the extension of some unrelated list, because a
```
tschatzl, line 144: Would it be possible to name these cases a/b/c instead of numbering them? The comments below also refer to these subcases as 1a-1c.

Cases a) and c) in this list seem to be in a different order than in the code, would be nice if they matched (just exchange).

I do not understand (a linguistic problem) why for case 3, `next_node` is the extension of some "unrelated" list, i.e. I do not understand the use of the adjective "unrelated" here, probably taking the word too literally.

The `result` to be popped must be in the same list as the list other threads pop from (e.g. in this case we always pop from the `_head` of the completed buffers list, racing with threads popping from the `_head` of the same list. There does not seem to be a way to compete with other, different lists here). Even if we got some element of a sub-list of the original list, it and its sublists seem still "related".

kimbarrett: The labeling is intentionally different.  The two groups are not congruent.  I thought
there was enough commentary there to make things clear, but apparently not.

kimbarrett:
After discussion with Thomas, I'm going to withdraw this PR and defer the fix
to JDK 19 rather than try to rush things.

